PR #29, opened 2013-10-23
fix for connection refused errors: go back to `getip.sh`

jakebiesinger:
@JavierJia 

As I mentioned on https://github.com/uci-cbcl/genomix/commit/0f31651729af55d028709904b58080768f5f4607#diff-fd9a88e53b0fd0f4e9040cf3157230eeR50 is seems there's a regression using the java inet lookup.  I played with it briefly but couldn't find a solution where the two agreed about the correct IP.

Since the `startCC.sh` and `startNC.sh` scripts use `getIP.sh`, I'm opting to keep it, at least until we can figure out a more general fix.


JavierJia:
OK, please go ahead and merge it 
LGTM

PR #40, opened 2013-10-25
Anbangx/debug bubble merge

anbangx:
A couple of changes in this feature branch:
1. Add output graphviz option to driver. Now you just now to specify -localGraphOutput and you can get the local graphviz output.
2. Add some optional log to bubblemerge to log the major vertex
3. For hg19 and dsm data, BubbleMerge pattern doesn't change the graph, though it doesn't cause any error. In order to double check that the graph really doesn't exist bubble, I need to extract subgraph by input major vertex(defined in bubble merge, remember?) as startSeed(you can start from this 'startSeed' and hop any times(set from parameter)). For bubble merge, we can only set the numOfHops as 2. It's good enough to check if it exists valid bubble. Now you just need to run "./extractSubgraph.sh hdfsInput localGraphOutput startSeed numOfHops" under the genomix/driver and then you can get the subgraph in the localGraphOutput directory you specify.
4. Fix a small bug in bubble merge.

@JavierJia @jakebiesinger @Nan-Zhang Please have a look at them. Thank you so much.


anbangx:
I will remove the text and graphviz from source control tonight. Sorry to keep them there.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/Node.java
```diff
@@ -84,6 +84,14 @@ public void set(EDGETYPE edgeType, VKmer kmer, ReadIdSet readIds) {
             this.kmer = kmer;
             this.readIds = readIds;
         }
+        
+        public String toString(){
+            StringBuilder sbuilder = new StringBuilder();
+            sbuilder.append('{');
+            sbuilder.append(kmer + " " + et);
```
JavierJia: If you want to take the benefit of stringbuilder, you should write

sbuilder.append(kmer.toString()).append(' ').append(et.toString()).

Otherwise, it is the same as string cat.

genomix/genomix-driver/extractSubgraph.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+#set -o pipefail
```
JavierJia: I don't know what's the meaning of this code . Can you explain this line to me ?

genomix/genomix-driver/extractSubgraph.sh
```diff
@@ -0,0 +1,19 @@
...
+    echo "for example:   $0 /user/anbangx/test-extract-graph ~/graph/ CCCTCACTCTCCCCTAACGCCCCTGCCTCCCAGACCCACGACTCACAGGAGGAGC 2"
+    exit 1
+fi
+
+OUTDIR=$2
+rm -rf $OUTDIR
+mkdir -p $OUTDIR
+
+# cd genomix/genomix-pregelix/
+target/appassembler/bin/genomix -kmerLength 55 -pipelineOrder EXTRACT_SUBGRAPH -hdfsInput $1 -localGraphOutput $OUTDIR -hdfsWorkPath /user/anbangx/genomix-out-extract-subgraph -startSeedToExtractSubgraph $3 -numHopsToExtractSubgraph $4
```
JavierJia: -hdfsWorkPath /user/anbangx/genomix-out-extract-subgraph
should be some other argument ?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/client/Client.java
```diff
@@ -69,6 +69,8 @@ private static Options prepareJob(String[] args, PregelixJob job) throws CmdLine
             job.getConfiguration().setInt(GenomixJobConf.GRAPH_CLEAN_MAX_ITERATIONS, options.numIteration);
         }
 
+        //        job.getConfiguration().set(ExtractSubgraphVertex.START_SEEDS, options.startSeeds);
```
JavierJia: What's the purpose of adding this comment out lines ?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/extractsubgraph/ExtractSubgraphVertex.java
```diff
@@ -0,0 +1,68 @@
...
+import edu.uci.ics.genomix.pregelix.io.VertexValueWritable;
+import edu.uci.ics.genomix.pregelix.io.VertexValueWritable.State;
+import edu.uci.ics.genomix.pregelix.io.message.MessageWritable;
+import edu.uci.ics.genomix.pregelix.operator.DeBruijnGraphCleanVertex;
+import edu.uci.ics.genomix.type.VKmer;
+import edu.uci.ics.pregelix.api.job.PregelixJob;
+
+public class ExtractSubgraphVertex extends DeBruijnGraphCleanVertex<VertexValueWritable, MessageWritable> {
+    private Set<VKmer> startSeeds = null;
+    private int numOfHops = -1;
```
JavierJia, line 25: Can you add some comments to explain the meaning of this numOfHops ?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/extractsubgraph/ExtractSubgraphVertex.java
```diff
@@ -0,0 +1,68 @@
...
+    public void markSelfAndBroadcast() {
+        VertexValueWritable vertex = getVertexValue();
+        vertex.setState(State.IS_MARK);
+        sendSettledMsgToAllNeighborNodes(vertex);
+    }
+
+    @Override
+    public void compute(Iterator<MessageWritable> msgIterator) throws Exception {
+        initVertex();
+        if (getSuperstep() == 1 && startSeeds.contains(getVertexId())) {
```
JavierJia: what's the different between these if /else branch ?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/extractsubgraph/ExtractSubgraphVertex.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+        voteToHalt();
+    }
+
+    public static PregelixJob getConfiguredJob(
+            GenomixJobConf conf,
+            Class<? extends DeBruijnGraphCleanVertex<? extends VertexValueWritable, ? extends MessageWritable>> vertexClass)
+            throws IOException {
+        PregelixJob job = DeBruijnGraphCleanVertex.getConfiguredJob(conf, vertexClass);
+        job.setVertexOutputFormatClass(ExtractSubgraphOutputFormat.class);
```
JavierJia, line 70: why we need this `job.setVertexOutputFormatClass` method ?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/format/ExtractSubgraphOutputFormat.java
```diff
@@ -0,0 +1,44 @@
...
+            BinaryVertexWriter<VKmer, VertexValueWritable, NullWritable> {
+        public BinaryLoadGraphVertexWriter(RecordWriter<VKmer, Node> lineRecordWriter) {
+            super(lineRecordWriter);
+        }
+
+        @Override
+        public void writeVertex(Vertex<VKmer, VertexValueWritable, NullWritable, ?> vertex) throws IOException,
+                InterruptedException {
+            byte state = (byte) (vertex.getVertexValue().getState() & State.VERTEX_MASK);
+            if (state == State.IS_MARK)
```
JavierJia: what's the meaning of "IS_MARK" ?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/io/VertexValueWritable.java
```diff
@@ -23,7 +23,8 @@
 
         public static final byte IS_NON = 0b1 << 6;
```
JavierJia: Can you add some comment to explain the meaning of each byte ?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/splitrepeat/SplitRepeatVertex.java
```diff
@@ -205,18 +235,30 @@ public void responseToRepeat(Iterator<SplitRepeatMessage> msgIterator) {
...
+            //     E -r2-> B -r1-> C -r3-> F
+            // B splits and delete his edge to A and C(B->A and B->C) in the 1th iteration
+            // in this iteration B also receives the message from C to delete edge B->C 
+            if(edgeMap.containsKey(deletedEdge.getKey()))
+                edgeMap.removeReadIdSubset(deletedEdge.getKey(), deletedEdge.getValue());
         }
     }
 
     @Override
     public void compute(Iterator<SplitRepeatMessage> msgIterator) {
```
JavierJia, line 259: Can you add some comment to explain what does this code do ?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
```diff
@@ -90,7 +90,10 @@ private static void loadLoggingFile() {
 
         @Option(name = "-localOutput", usage = "Local directory where the final step's output will be saved", required = false)
         private String localOutput;
-
+        
+        @Option(name = "-localGraphOutput", usage = "Local directory where the final step's graph output will be saved", required = false)
```
jakebiesinger: This is a good idea, but why not let the graph be dumped at any point along the way?  It can be a regular pipeline pattern, just like STATS or DUMP_FASTA. STATS currently dumps to HDFS but there's no reason it shouldn't have local output, IMHO.  I think DUMP_FASTA does local output to something like genome.fasta.

What if our "-localOutput" folder contained several items instead of just the last steps contents? By default, there is no localOutput, but let's pretend it was `genomix-results`.  What I'm envisioning is something like: `-pipelineOrder BUILD,STATS,MERGE,STATS,BUBBLE,STATS,SHOW_GRAPH,DUMP_FASTA` leading to _local_ output like:

```
genomix-results/
|-- 01-BUILD-STATS
|   |-- coverage-hist.png
|   |-- kmerLength-hist.png
|   `-- stats.txt
|-- 02-MERGE-STATS
|   |-- coverage-hist.png
|   |-- kmerLength-hist.png
|   `-- stats.txt
|-- 03-BUBBLE-DUMP_FASTA
|   `-- genome.fasta
|-- 03-BUBBLE-SHOW_GRAPH
|   |-- graph.dot
|   `-- graph.svg
`-- 03-BUBBLE-STATS
    |-- coverage-hist.png
    |-- kmerLength-hist.png
    `-- stats.txt
```

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
```diff
@@ -146,6 +149,12 @@ private static void loadLoggingFile() {
         
         @Option(name = "-minScaffoldingVertexMinLength", usage = "The minimum vertex length that can be the head of scaffolding", required = false)
         private int minScaffoldingVertexMinLength = -1;
+        
+        @Option(name = "-startSeedToExtractSubgraph", usage = "The minimum vertex length that can be the head of scaffolding", required = false)
```
jakebiesinger: a more consistent name would be extractSubgraph_startSeed

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
```diff
@@ -146,6 +149,12 @@ private static void loadLoggingFile() {
...
+        private String startSeedToExtractSubgraph;
+        
+        @Option(name = "-numHopsToExtractSubgraph", usage = "The minimum vertex length that can be the head of scaffolding", required = false)
```
jakebiesinger: extractSubgraph_numHops

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
```diff
@@ -188,6 +197,7 @@ private static void loadLoggingFile() {
         SPLIT_REPEAT,
         DUMP_FASTA,
         CHECK_SYMMETRY,
+        EXTRACT_SUBGRAPH,
```
jakebiesinger: Will this step be used for anything other than plotting?  As it stands, you're extracting an invalid subgraph, right?  Meaning you have dangling edges?

If we can't think of a use for this step outside of plotting, then let's roll the PLOT_GRAPH and this step into one step with the startSeed and numHops as optional values (so the SHOW_GRAPH would run on the entire graph if you don't specify the seed).

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/minicluster/GenerateGraphViz.java
```diff
@@ -154,4 +154,62 @@ public static String getColor(EDGETYPE et) {
                 throw new IllegalStateException("Invalid input Edge Type!!!");
         }
     }
+    
+    /**
+     * simple version of graphviz, only print the node
+     */
+    public static void convertGraphCleanOutputToSimpleNode(String srcDir, String destDir) throws Exception {
```
jakebiesinger: It seems like this function is a duplicate of previous work; can you just add a boolean or numeric parameter for verbosity?  For example, level 0 would be like an undirected graph w/out any labels; 1 would be a graph with edge types but no other labels; 2 would include the kmers or some ID of the graph; 3 would include details like 5' and ~5' readids, internal kmers, etc.  

Then it's just one function for all the plotting.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/Node.java
jakebiesinger: string concatenation is O(N^2) in java... is that right? Do we care?

genomix/genomix-driver/extractSubgraph.sh
jakebiesinger: this is a bash-ism which causes the whole script to fail if any step in `a | pipeline of | commands like | this` fails.  But it's commented out right now so it doesn't do anything.

genomix/genomix-driver/extractSubgraph.sh
jakebiesinger: also, this assumes it is run from the hyracks root... that isn't likely to be the case...

jakebiesinger: Or at least from the genomix-driver subdirectory.  If you want a script like this, put it in `genomix-driver/src/main/resources/scripts`. Then it will show up in `target/appassembler`.

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
```diff
@@ -144,6 +147,10 @@ private void addStep(GenomixJobConf conf, Patterns step) throws Exception {
                 queuePregelixJob(SymmetryCheckerVertex.getConfiguredJob(conf, SymmetryCheckerVertex.class));
                 curOutput = prevOutput; // use previous job's output
                 break;
+            case EXTRACT_SUBGRAPH:
+                queuePregelixJob(ExtractSubgraphVertex.getConfiguredJob(conf, ExtractSubgraphVertex.class));
+//                curOutput = prevOutput; // use previous job's output
```
jakebiesinger: I think you better uncomment this... unless you want the next job to run on whatever output you're generating (maybe that exactly what you want?)

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
```diff
@@ -193,6 +200,9 @@ private void queuePregelixJob(PregelixJob job) {
         if (job.getClass().equals(SymmetryCheckerVertex.class)) {
             job.setVertexOutputFormatClass(CheckerOutputFormat.class);
         }
+        if (job.getClass().equals(ExtractSubgraphVertex.class)) {
```
jakebiesinger, line 223: All of these special cases should be rolled into the `getConfiguredJob` and this function should be removed.  Its only purpose was to handle -followsGraphBuild and P2 complexity stuff but that's all handled properly between the new input format and the `getConfiguredJob`.

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
```diff
@@ -293,10 +312,14 @@ public void runGenomix(GenomixJobConf conf) throws NumberFormatException, Hyrack
     }
 
     public static void main(String[] args) throws CmdLineException, NumberFormatException, HyracksException, Exception {
-        String[] myArgs = { "-runLocal", "true", "-kmerLength", "55",
```
jakebiesinger, line 300: Please don't commit changes to this stuff. In fact, let's go ahead and strip it out.  You can have all the local changes you want, but we shouldn't let these kinds of changes touch our repo.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/client/Client.java
jakebiesinger: What's the purpose of this class?  Do you actually use it?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/format/ExtractSubgraphOutputFormat.java
```diff
@@ -0,0 +1,44 @@
...
+            throws IOException, InterruptedException {
+        @SuppressWarnings("unchecked")
+        RecordWriter<VKmer, Node> recordWriter = binaryOutputFormat.getRecordWriter(context);
+        return new BinaryLoadGraphVertexWriter(recordWriter);
+    }
+
+    /**
+     * Simple VertexWriter that supports {@link BinaryLoadGraphVertex}
+     */
+    public static class BinaryLoadGraphVertexWriter extends
```
jakebiesinger, line 30: I thought we could simplify most of these output formats since they inherit functionality?  Is this as simple as this one can get?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/io/VertexValueWritable.java
```diff
@@ -23,7 +23,8 @@
...
         public static final byte IS_ERROR = 0b1 << 6;
-
+        public static final byte IS_MARK = 0b1 << 6;
```
jakebiesinger: It seems these are using the same bits as placeholders.... I sense some trouble here...

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/DeBruijnGraphCleanVertex.java
```diff
@@ -90,6 +90,7 @@ public void checkDebug() {
         }
 
         verbose = false;
+        verbose |= debug;
```
jakebiesinger: so all nodes are verbose if debug is true???

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/SimpleBubbleMergeVertex.java
```diff
@@ -12,6 +13,7 @@
 import edu.uci.ics.genomix.pregelix.io.message.BubbleMergeMessage;
 import edu.uci.ics.genomix.pregelix.operator.DeBruijnGraphCleanVertex;
 import edu.uci.ics.genomix.pregelix.operator.aggregator.StatisticsAggregator;
+import edu.uci.ics.genomix.pregelix.operator.pathmerge.P4ForPathMergeVertex;
```
jakebiesinger: what do you need this import for?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/SimpleBubbleMergeVertex.java
```diff
@@ -73,6 +78,9 @@ public void sendBubbleAndMajorVertexMsgToMinorVertex() {
             outgoingMsg.setMinorToBubbleEdgetype(forwardNeighbor.et.mirror());
             minorVertexId = forwardNeighbor.kmer;
         }
+        if(verbose){
```
jakebiesinger: don't forget to ctrl+shift+f

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/splitrepeat/SplitRepeatVertex.java
```diff
@@ -41,7 +41,7 @@
     private static final Logger LOG = Logger.getLogger(SplitRepeatVertex.class.getName());
 
     public static final int NUM_LETTERS_TO_APPEND = 3;
-    private static long randSeed = 1; //static for save memory
```
jakebiesinger: Umm??? We've seen all these changes in your other branch.... That makes reviewing very difficult.

When you submit a pull request, you can specify a base branch (rather than the default genomix/fulstack_genomix). You should have made this branch based on fullstack but if you absolutely need it based on another branch, you should set it up that way so these changes aren't reviewed twice.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
anbangx: I agree with you!

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
anbangx: Done!

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
anbangx: Done!

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
anbangx: This step is only used for plotting. The goal for it is to check the graph structure.

Yes, it's invalid subgraph because it has dangling edges. 

How about PLOT_SUBGRAPH? And if we can't specify the seed, we output nothing. Because we have another option -localGraphOutput to output the entire graph.

jakebiesinger: Why output nothing?  If I requested you to plot the graph, I want you to plot the graph! No matter how big it is!

The `-localGraphOutput` option can be removed if we use the `-localOutput` directory as a container for many kinds of output, as I mentioned above.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
jakebiesinger: Umm?  I don't see the change...

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
jakebiesinger: Nor here...

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/minicluster/GenerateGraphViz.java
anbangx: Agree! 

Done. Now you can pick different details by specifying -extractSubgraph_graphType.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/Node.java
anbangx: Done! @JavierJia Could you explain more details about "string concatenation" and "stringbuilder"?

genomix/genomix-driver/extractSubgraph.sh
anbangx: It's exactly like what Jake said.

genomix/genomix-driver/extractSubgraph.sh
anbangx: remove the hdfsWorkPath arg and move it to src/main/resources/scripts

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
anbangx: Actually if you uncomment it, the output is not correct.

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
anbangx, line 300: Got it

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/client/Client.java
anbangx: This class is used to run job only in pregel package. I think I can REMOVE it. How do you think? @JavierJia  @jakebiesinger

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/extractsubgraph/ExtractSubgraphVertex.java
anbangx, line 25: Add it~

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/extractsubgraph/ExtractSubgraphVertex.java
anbangx: Simplify it!

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/extractsubgraph/ExtractSubgraphVertex.java
anbangx, line 70: Because the output condition of ExtractSubgraph is different with other patterns. If the vertex is what I want to extract, I will mark IS_MARK its state and check this condition when output.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/format/ExtractSubgraphOutputFormat.java
anbangx, line 30: No. The vertex format of this output format should be <VKmer, VertexWritable> instead of <VKmer, NodeWritable>, because we need check the condition when trying to output. The vertices which we try to extract will be marked as IS_MARK and check it when output

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/format/ExtractSubgraphOutputFormat.java
anbangx: explain before.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/io/VertexValueWritable.java
anbangx: Add it~

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/io/VertexValueWritable.java
anbangx: Use for different functionalities. They can be reused.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/minicluster/GenerateGraphViz.java
jakebiesinger: how about `-plotSubgraph_verbosity`?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/extractsubgraph/ExtractSubgraphVertex.java
jakebiesinger, line 70: I think IS_MARK is a poor name.  How about `SHOULD_KEEP` or `KEEP_NODE`?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/client/Client.java
jakebiesinger: yeah, remove it.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/SimpleBubbleMergeVertex.java
anbangx: Done/

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/SimpleBubbleMergeVertex.java
anbangx: Done

genomix/genomix-driver/extractSubgraph.sh
jakebiesinger: I don't see the change...

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/splitrepeat/SplitRepeatVertex.java
anbangx: Got it!

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/splitrepeat/SplitRepeatVertex.java
anbangx, line 259: The explanation is on top of the class!

anbangx:
Some refactors based on code review.

@JavierJia  @jakebiesinger  @Nan-Zhang  Check again. Thanks.

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
jakebiesinger: commented out, the next step will run on the subgraph with dangling edges.  For example, try running `EXTRACT_SUBGRAPH` followed by `MERGE`... it will certainly fail!

Once you replace `EXTRACT_SUBGRAPH` with `PLOT_SUBGRAPH`, you'll want to have the output be the original graph and will need to uncomment this line.

jakebiesinger:
I still see many comments unaddressed.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/io/VertexValueWritable.java
jakebiesinger: I guess so... It's pretty ugly to join them together in the same class though. You could restrict the overlap via inheritance (so different derived classes can use the same bits).  It's just code smell-- I see two flags with the same bits and have a knee-jerk reaction that this is a bad idea.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
```diff
@@ -32,6 +32,7 @@
 import org.kohsuke.args4j.CmdLineParser;
 import org.kohsuke.args4j.Option;
 
+import edu.uci.ics.genomix.minicluster.GenerateGraphViz.GRAPG_TYPE;
```
jakebiesinger: Spelling error?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
```diff
@@ -91,6 +92,9 @@ private static void loadLoggingFile() {
         @Option(name = "-localOutput", usage = "Local directory where the final step's output will be saved", required = false)
         private String localOutput;
 
+        @Option(name = "-localGraphOutput", usage = "Local directory where the final step's graph output will be saved", required = false)
```
jakebiesinger: Didn't we discuss having a single localOutput directory so you can use the previous step's directory name as the local output dir? like `03-MERGE-PLOT_SUBGRAPH`?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
```diff
@@ -133,20 +137,29 @@ private static void loadLoggingFile() {
...
         @Option(name = "-minScaffoldingVertexMinLength", usage = "The minimum vertex length that can be the head of scaffolding", required = false)
         private int minScaffoldingVertexMinLength = -1;
 
+        @Option(name = "-extractSubgraph_startSeed", usage = "The minimum vertex length that can be the head of scaffolding", required = false)
+        private String extractSubgraph_startSeed;
+
+        @Option(name = "-extractSubgraph_numHops", usage = "The minimum vertex length that can be the head of scaffolding", required = false)
+        private int extractSubgraph_numHops = -1;
+        
+        @Option(name = "-extractSubgraph_graphType", usage = "Specify the order of the graph cleaning process", required = false)
```
jakebiesinger: Still like `-extractSubgraph_verbosity` better... and you're not specifying an order at all here, are you? 

Plus, the usage message should indicate what will be shown in the graph, as I mentioned in my previous comment.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
```diff
@@ -259,6 +274,9 @@ public static void verifyPatterns(Patterns[] patterns) {
     public static final String SCAFFOLDING_MAX_TRAVERSAL_LENGTH = "scaffolding.max.traveral.length";
     public static final String SCAFFOLDING_VERTEX_MIN_COVERAGE = "scaffolding.vertex.min.coverage";
     public static final String SCAFFOLDING_VERTEX_MIN_LENGTH = "scaffolding.vertex.min.length";
```
jakebiesinger, line 270: Maybe we should make all the genomix-specific options have `genomix.` as the first element... that will cause the conf file to be more readable.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
```diff
@@ -373,9 +391,6 @@ private void fillMissingDefaults() {
         if (getLong(RANDOM_RANDSEED, -1) == -1)
```
jakebiesinger: just `RANDOM_SEED`

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
```diff
@@ -404,6 +419,16 @@ private void fillMissingDefaults() {
         if (get(PIPELINE_ORDER) == null) {
             set(PIPELINE_ORDER, Patterns.stringFromArray(DEFAULT_PIPELINE_ORDER));
         }
+        if (get(EXTRACT_SUBGRAPH_GRAPH_TYPE) == null)
+            set(EXTRACT_SUBGRAPH_GRAPH_TYPE, GRAPG_TYPE.DIRECTED_GRAPH_WITH_SIMPLELABEL_AND_EDGETYPE.toString());
```
jakebiesinger: Holy smokeage... the user is supposed to write `-extractSubgraph_graphType DIRECTED_GRAPH_WITH_SIMPLELABEL_AND_EDGETYPE` ???

Let's just make it numeric!

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/minicluster/GenerateGraphViz.java
```diff
@@ -11,67 +12,31 @@
...
 import edu.uci.ics.genomix.type.ReadIdSet;
 import edu.uci.ics.genomix.type.VKmer;
 
 //TODO by Jianfeng: move this to script
 public class GenerateGraphViz {
 
+    private static long count = 0;
+    private static HashMap<String, Long> map = new HashMap<String, Long>();
+
+    public enum GRAPG_TYPE {
```
jakebiesinger: spelling error

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/minicluster/GenerateGraphViz.java
```diff
@@ -11,67 +12,31 @@
...
+        
+        UNDIRECTED_GRAPH_WITHOUT_LABELS,
```
jakebiesinger: Sorry but I think this is one place where a number is more helpful than an enum... or at least make it backed by a number.  You expect the user to write these insanely long names on the command line?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/minicluster/GenerateGraphViz.java
```diff
@@ -93,18 +58,37 @@ public static void convertGraphCleanOutputToGraphViz(String srcDir, String destD
                 if (key == null) {
                     break;
                 }
-                outputNode += key.toString();
+                switch (graphType) {
```
jakebiesinger, line 85: Numeric values would allow later flexibility... for example, you could say "graphType > UNDIRECTED_GRAPH_WITHOUT_LABELS" and do the things for those several cases.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/minicluster/GenerateGraphViz.java
```diff
@@ -154,4 +163,62 @@ public static String getColor(EDGETYPE et) {
                 throw new IllegalStateException("Invalid input Edge Type!!!");
         }
     }
+
+    //    /**
+    //     * simple version of graphviz, only print the node
```
jakebiesinger: Don't need to leave commented out code in place.

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
```diff
@@ -144,6 +148,10 @@ private void addStep(GenomixJobConf conf, Patterns step) throws Exception {
                 queuePregelixJob(SymmetryCheckerVertex.getConfiguredJob(conf, SymmetryCheckerVertex.class));
                 curOutput = prevOutput; // use previous job's output
                 break;
+            case EXTRACT_SUBGRAPH:
```
jakebiesinger: still waiting for PLOT_SUBGRAPH refactor.

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
```diff
@@ -286,6 +297,16 @@ public void runGenomix(GenomixJobConf conf) throws NumberFormatException, Hyrack
 
         if (conf.get(GenomixJobConf.LOCAL_OUTPUT_DIR) != null)
             GenomixClusterManager.copyBinToLocal(conf, curOutput, conf.get(GenomixJobConf.LOCAL_OUTPUT_DIR));
+        if (conf.get(GenomixJobConf.LOCAL_GRAPH_OUTPUT_DIR) != null) {
```
jakebiesinger: This should go away as you should be able to plot at any step.

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
```diff
@@ -293,10 +314,14 @@ public void runGenomix(GenomixJobConf conf) throws NumberFormatException, Hyrack
...
     public static void main(String[] args) throws CmdLineException, NumberFormatException, HyracksException, Exception {
-        String[] myArgs = { "-runLocal", "true", "-kmerLength", "55",
+        String[] myArgs = {
+                "-kmerLength",
+                "55",
+                //                        "-runLocal", "true", 
                 //                        "-saveIntermediateResults", "true",
                 //                        "-localInput", "../genomix-pregelix/data/input/reads/synthetic/",
-                "-localInput", "tail600000",
+                "-localInput",
```
jakebiesinger: still see these erroneous commits... just blow away this whole section.

genomix/genomix-driver/src/main/resources/scripts/extractSubgraph.sh
```diff
@@ -0,0 +1,21 @@
...
+
+OUTDIR=$2
+rm -rf $OUTDIR
+mkdir -p $OUTDIR
+
+# cd genomix/genomix-pregelix/
+# target/appassembler/bin/genomix -kmerLength 55 -pipelineOrder EXTRACT_SUBGRAPH -hdfsInput $1 -localGraphOutput $OUTDIR -hdfsWorkPath /user/anbangx/genomix-out-extract-subgraph -startSeedToExtractSubgraph $3 -numHopsToExtractSubgraph $4
+# cd ../..
+
+bin/genomix -kmerLength 55 -pipelineOrder EXTRACT_SUBGRAPH -hdfsInput $1 -localGraphOutput $OUTDIR -startSeedToExtractSubgraph $3 -numHopsToExtractSubgraph $4
```
jakebiesinger: look at the other scripts-- they resolve app.home themselves using `pwd` and parent dirs.  This script won't work as-is from the bin directory; the other will.

genomix/genomix-pregelix/data/TestSet/BubbleMerge/SimpleLongReads/graphviz.svg
```diff
@@ -0,0 +1,113 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
jakebiesinger, line 1: Please son't commit the svg files...

anbangx, line 1: Different branches..... I changed in other branch

On Fri, Oct 25, 2013 at 12:18 PM, Jake Biesinger
notifications@github.comwrote:

> In
> genomix/genomix-pregelix/data/TestSet/BubbleMerge/SimpleLongReads/graphviz.svg:
> 
> > @@ -0,0 +1,113 @@
> > +<?xml version="1.0" encoding="UTF-8" standalone="no"?>
> 
> Please son't commit the svg files...
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/uci-cbcl/genomix/pull/40/files#r7225540
> .

## 

Best Regards,

Anbang Xu

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/DeBruijnGraphCleanVertex.java
jakebiesinger: this still seems incorrect to me...

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
anbangx: Default...

jakebiesinger: I know this is the default value... but I'm pointing out the `TERRIBLY_LONG_NAME_THAT_THE_USER_WOULD_HAVE_TO_WRITE_TO_CHANGE_THE_DEFAULT` and opting instead for a simple int which you can convert into an enum.

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
anbangx, line 223: You want to do it or you want me to do it?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/Node.java
JavierJia: The string concatenation will create a new string. the stringbuilder will not. 
I just want to keep consistent with the code logic, if we decide to use stringbuilder ,then we should use it correct.

genomix/genomix-driver/extractSubgraph.sh
JavierJia: Good to know. Thanks.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/DeBruijnGraphCleanVertex.java
anbangx: I want to make -debugKmers "" works.

anbangx:
1. Rename to PLOT_SUBGRAPH and provide simple numeric value for users to choose.
2. Resolve app.home themselves using pwd and parent dirs in script

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/DeBruijnGraphCleanVertex.java
jakebiesinger: But this will cause literally every node to log when you specify _any_
debug kmers.

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
jakebiesinger, line 223: I'll do it tonight.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/DeBruijnGraphCleanVertex.java
anbangx: Right. So if it is for debug purpose and trying to trace each vertex, we can directly use if(debug). Correct?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
anbangx, line 270: Agree! You want to do it or you want me to do it?

anbangx:
Two refactors:
1. Make PLOT_SUBGRAPH as a pregel job, you can dump the subgraph in any step, and it will output to the same dir as the previous job. Also, if you run in the last step of the whole pipeline, you can specify the -localInput option to get it.
2. Provide numerical value for users to choose among different GRAPH_TYPE.

@jakebiesinger @JavierJia @Nan-Zhang Have a look on them.

jakebiesinger:
Were you successful at making the subplot a separate job?  The commit message sounds like it's a work in progress (meaning you're not done).

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
jakebiesinger, line 270: Let's make it a separate pull request.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/DeBruijnGraphCleanVertex.java
jakebiesinger: That's right.  Just `if(debug)` should work if you specify `-debugKmers ""`.

## 

Jake Biesinger
Graduate Student
Xie Lab, UC Irvine

anbangx:
Hi ALL,

I fixed the Remove_Lowcoverage in this branch. @jakebiesinger Please try this one. 

The previous bug is in a static value.

jakebiesinger:
Yeah looks like the case I showed you before has gone away.  LGTM.

anbangx:
Last refactor in this feature branch. @JavierJia @Nan-Zhang @jakebiesinger  Have a look and then I will merge with main branch.
1. Provide numerical value for users but deal with Enum under the hook

anbangx:
Done!

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
jakebiesinger, line 223: fixed in #47.

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
jakebiesinger, line 300: Fixed in #47.

jakebiesinger:
I just noticed you already closed and merged this branch. My last comment here is still in place if you ask me-- neither `DEFAULT_PIPELINE_ORDER` nor `DEFAULT_GRAPH_TYPE` belong here.  Let's put them in `fillMissingDefaults`.

anbangx:
DONE!

On Thu, Oct 31, 2013 at 11:45 AM, Jake Biesinger
notifications@github.comwrote:

> I just noticed you already closed and merged this branch. My last comment
> here is still in place if you ask me-- neither DEFAULT_PIPELINE_ORDER nor
> DEFAULT_GRAPH_TYPE belong here. Let's put them in fillMissingDefaults.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/uci-cbcl/genomix/pull/40#issuecomment-27514077
> .

## 

Best Regards,

Anbang Xu

jakebiesinger:
Where is it implemented?

anbangx:
main branch

On Thu, Oct 31, 2013 at 12:37 PM, Jake Biesinger
notifications@github.comwrote:

> Where is it implemented?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/uci-cbcl/genomix/pull/40#issuecomment-27519007
> .

## 

Best Regards,

Anbang Xu

PR #47, opened 2013-10-31
Simplify and robustify cluster startup and config

jakebiesinger:

These changes allow us to run a hyracks cluster at the same time as a pregelix cluster and greatly simplify the user's interaction with the `*.properties` files.
## Config file changes

Previously, a single config file was used to specify ports, working dirs, and logs for the currently active cluster.  Rather than force the user to maintain two different config files (one for hyracks and one for pregelix), we now use a single, simplified config file but _automatically generate_ the two cluster-type specific configuration files required.

The simplified file is still under `conf/cluster.properties` though we may want to change the name now that its values are incompatible with other cluster.properties we've seen so far.  This file contains:
- `$WORKPATH`: the root for temporary jars and logs; also the working directory for the cc's and nc's. The four working directories for the CC's and NC's are now `$WORKPATH/hyracks/cc` ... `$WORKPATH/pregelix/nc`.
- `$DISKS`: a comma-separated list of folders corresponding to different physical disks.  This is _like_ `IO_DIRS` and `store` but is not the same-- the number of disks has no relationship to the number of partitions/workers/threads in use. Rather, the user-prescribed `-threadsPerMachine` will round-robin over `$DISKS` such that multiple partitions may be active on the same disk (with low `-threadsPerMachine` and high `$DISKS`, some disks may not have any partitions on them at all).  `IO_DIRS` (and `store` for pregelix) are now placed under each of the `$DISKS` as `$DISKS[0]/hyracks/io_dir-0` through `$DISKS[maxThreads % len($DISKS)]/hyracks/io_dir-$maxThreads`.
- Standard ports for the CC (client, cluster and http), plus CC_DEBUG and NC_DEBUG (which are used to allow Eclipse connections to the CC and NC).  Notice there are separate entries now for HYRACKS and PREGELIX.
- `NC_MAX_MEMORY`: self-explanatory, I think...

The corresponding config files are not generated until the `startCluster.sh` script is called. Normally, this is done when `genomix` is run, but you can also start a cluster independently and specify `-useExistingCluster`.
## Driver changes
- `-threadsPerMachine <int>` (defaults to 1)
- `-useExistingCluster` will skip startup and shutdown of the CC's and NC's. The existing cluster must still use the ports specified in `conf/cluster.properties`.
- removed `-followsGraphBuild` (not necessary anymore, thanks to Anbang's refactoring the pregelix input formats)
- removed `-clusterWaitTime` as the wait is not handled inside of the `startCluster.sh` scripts.
- both pregelix and hyracks CC's and NC's are started at the very beginning rather than being started/stopped during the run.
- we now detect startup errors in the `startcc.sh` and `startnc.sh` scripts-- If there is an error (e.g., `Address already in use`), it will be printed to the command line and driver execution will stop immediately.
## Logging changes
- `logging.properties` has been broken into two separate files, `driver.logging.properties` and `worker.logging.properties.  The former is used by the driver (where you don't want to see INFO message from hadoop) whereas the latter is for the CC and NC (where finer-grained messages are more helpful). 
- I've also fixed the problems we had where you had to specify the absolute path of the logging file for the NC and CC to pick them up.  
- We also no longer need the ugly static initialization block `loadLoggingFile`-- changes to the `pom.xml` fixed that.
- I still don't have a solution for the driver `%h/genomix-%u.log` file we use often when debugging-- it will still be written to your home directory. 
## Maven changes
- I've stripped away `src/main/resources/conf` and `src/main/resources/scripts` everywhere except for `genomix-driver` to reduce confusion surrounding very similarly-named scripts that are now out of date.
- A lot more of the configuration occurs in the driver's `pom.xml`, including setting the logging files and allowing the CC and NC's each have their own `app.home` directory while still sharing a single `lib`.
- I fixed a bug where the `binary-assembly` folder would sometimes be missing scripts.
- using `appassembler` as your base directory is now deprecated in favor of using `genomix-driver-2.10-SNAPSHOT` (notice the lack of `binary-assembly`-- I removed that suffix).


jakebiesinger:
One kind of big change is that `-runLocal` now starts a miniDFS and miniMR cluster.  We don't actually need either and there's a lot of overhead with them (and a TON of logging) but maybe we believe our test cases more if they're run against a proper HDFS instance?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/minicluster/GenomixClusterManager.java
```diff
@@ -278,81 +282,36 @@ private static String getAbsolutePathFromURI(String URI) {
...
-        if (shutdownHooks.containsKey(clusterType))
-            throw new IllegalArgumentException("Already specified a hook for shutting down a " + clusterType
-                    + " cluster! (Try removing the existing hook first?)");
-        Thread hook = new Thread() {
+    private void addClusterShutdownHook() {
+        if (shutdownHook != null) {
+            throw new IllegalArgumentException(
+                    "Already specified a hook for shutting down the cluster (Try removing the existing hook first?)");
+        }
+        shutdownHook = new Thread() {
```
JavierJia, line 290: I was always curious what's the purpose of this `shutdownHook` ?

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
```diff
@@ -231,39 +205,45 @@ private void flushPendingJobs(GenomixJobConf conf) throws Exception {
...
         stepNum = 0;
         runLocal = Boolean.parseBoolean(conf.get(GenomixJobConf.RUN_LOCAL));
+
         manager = new GenomixClusterManager(runLocal, conf);
-        manager.stopCluster(ClusterType.HYRACKS); // shut down any existing NCs and CCs
+        if (!Boolean.parseBoolean(conf.get(GenomixJobConf.USE_EXISTING_CLUSTER))) {
+            manager.stopCluster(); // shut down any existing NCs and CCs
+            manager.startCluster();
+        }
+        if (runLocal) {
```
JavierJia, line 272: So time `runlocal` just mean that start a local cluster ? Don't mean that the jvm stuff any more ?

JavierJia:
Awesome work ! :+1:

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/minicluster/GenomixClusterManager.java
jakebiesinger, line 290: If the user hits `ctrl+c` to interrupt genomix, we would normally leave the cluster running which may be problematic for subsequent runs. The shutdown hook runs just before the JVM closes and in this case shuts down anything we left running.

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
jakebiesinger, line 272: yeah... there are really four options we _could_ have, and we currently make available three:
- `-runLocal`: create _within-JVM_ NC's and CC's, all talking over `localhost`.  This is the only way to debug work within an NC without attaching to a remote process. This will also create a miniMR and miniDFS cluster.
- `-runLocalNoDFS` (not implemented): same as `-runLocal` but doesn't create any MR or DFS clusters.  This mode is a lot easier to work with (anything in "HDFS" is actually on the local, raw filesystem) but maybe(?) isn't the preferred mode for testing?
- default mode (nothing specified on the command line): starts separate CC's and NC's for hyracks and pregelix at startup time; always shuts them down (whether b/c of early termination or pipeline completion).
- `-useExistingCluster`: assumes you've already run the `startCluster.sh` script and won't start up or shutdown anything.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/minicluster/GenomixClusterManager.java
JavierJia, line 290: Cool :+1:

jakebiesinger:
@JavierJia travis is running everything fine except your HyracksVSHadoop test!  That test runs fine on my local machine :frowning: It's just showing "job failed".  I'll test a commit that doesn't squelch the hadoop output.

JavierJia:
Let me check it ~

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/minicluster/DriverUtils.java
```diff
@@ -62,42 +62,28 @@ public static String getIP(String hostName) throws IOException, InterruptedExcep
...
-        } else {
-            if (conf.get(GenomixJobConf.IP_ADDRESS) == null) {
-                conf.set(GenomixJobConf.IP_ADDRESS, getIP("localhost"));
-            }
-            if (conf.getInt(GenomixJobConf.PORT, -1) == -1) {
-                conf.set(GenomixJobConf.PORT, CCProperties.getProperty("CC_CLIENTPORT"));
-            }
-        }
 
-        if (conf.get(GenomixJobConf.FRAME_SIZE) == null)
```
JavierJia, line 86: @jakebiesinger  the problem in my observation is that the genomix-hyracks can not get the FRAME_SIZE and the FRAME_LIMIT.
When it run conf.get(GenomixJobConf.FRAME_SIZE), it returned an "null". Any idea ?

jakebiesinger, line 86: Is this you running locally? Or on Travis? Sounds like the path to the conf
file is incorrect or out hasn't been generated yet.

If local, make sure you do mvn package first. If on Travis, echo the path
first.

JavierJia, line 86: Run locally on my machine. 
And if I use junit test, every works fine. If I use `mvn test` it will fail.

JavierJia, line 86: It reports like this

```
Caused by: java.lang.NumberFormatException: null
    at java.lang.Integer.parseInt(Integer.java:454)
    at java.lang.Integer.parseInt(Integer.java:527)
    at edu.uci.ics.genomix.hyracks.graph.job.JobGen.initGenomixConfiguration(JobGen.java:97)
    at edu.uci.ics.genomix.hyracks.graph.job.JobGenBuildBrujinGraph.initGenomixConfiguration(JobGenBuildBrujinGraph.java:72)
    at edu.uci.ics.genomix.hyracks.graph.job.JobGen.<init>(JobGen.java:79)
```

But I checked the code they should be set properly in the test code.

JavierJia, line 86: I fixed the hyrack related problem and pushed the commit.
But there is still problem while run hadoop driver. 

```
java.io.IOException: Job failed!
    at org.apache.hadoop.mapred.JobClient.runJob(JobClient.java:1252)                                          
    at edu.uci.ics.genomix.hadoop.contrailgraphbuilding.GenomixHadoopDriver.run(GenomixHadoopDriver.java:83)
    at edu.uci.ics.genomix.driver.comparation.HyrackVSHadoopTest.TestEachFile(HyrackVSHadoopTest.java:125)
```

Nan-Zhang:
@jakebiesinger  this is a pretty cool change for running genomix more conveniently
I have some questions about theses changes:
1. now we removed -followsGraphbuild, if I want run pathmerging directly (based on my graphbuilding results which I already have), what should I do? should I move the graphbuilding's results to some specific dir?
2. the appassembler is deprecated, if I want to run genomix-driver, I should go to genomix-driver-0.2.10-SNAPSHOT folder, then type "bin/genomix blablablabla", is my thought right? does generating the genomix-driver-0.2.10-SNAPSHOT have same rule with generating appassembler, which both copy the conf and scripts from src/main/resource to genomix-driver-0.2.10-SNAPSHOT? And I don't found the the binary-assembly, so we removed it, so what's original meaning for this binary-assembly existence?

jakebiesinger:
@Nan-Zhang 1. will just work-- you don't have to do anything extra whether it's after graphbuild or not.  2. yes, don't use the appassembler and instead cd to the snapshot directory (or just use the script-- you shouldn't have to cd at all). 3. yes, I removed the "binary-assembly" suffix as it just confuses the user (already too much to have all this SNAPSHOT business).

.travis.yml
```diff
@@ -19,6 +19,6 @@ script:
     - cd genomix
     - mvn package -q -e -Djava.util.logging.config.file=genomix-driver/src/main/resources/conf/logging.properties
     - ls -t */target/surefire-reports/*.txt | xargs tail -n +1  # show test reports in the order they ran
-    - cd genomix-driver/target/genomix-driver-0.2.10-SNAPSHOT-binary-assembly
+    - cd genomix-driver/target/genomix-driver-0.2.10-SNAPSHOT
     - bin/genomix -kmerLength 55 -pipelineOrder BUILD_HYRACKS,MERGE,TIP_REMOVE,MERGE,BUBBLE,MERGE -localInput ../../src/test/resources/data/sequence/pathmerge/hg19.chr18.skip12K.first1K.1Kreads/
```
anbangx, line 25: Only check this pipelineOrder is not enough. How about running the default pipeline: BUILD, MERGE, LOW_MERGE, TIP_REMOVE, MERGE, BUBBLE, SPLIT_REPEAT, MERGE.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/minicluster/DriverUtils.java
```diff
@@ -62,42 +62,28 @@ public static String getIP(String hostName) throws IOException, InterruptedExcep
...
-            conf.set(GenomixJobConf.FRAME_SIZE, CCProperties.getProperty("FRAME_SIZE"));
-        if (conf.get(GenomixJobConf.FRAME_LIMIT) == null)
-            conf.set(GenomixJobConf.FRAME_LIMIT, CCProperties.getProperty("FRAME_LIMIT"));
-        if (conf.get(GenomixJobConf.HYRACKS_IO_DIRS) == null)
-            conf.set(GenomixJobConf.HYRACKS_IO_DIRS, CCProperties.getProperty("IO_DIRS"));
-        if (conf.get(GenomixJobConf.HYRACKS_SLAVES) == null) {
+        for (String prop : clusterProperties.stringPropertyNames()) {
```
anbangx, line 71: Where do you set genomix-hyracks.FRAME_SIZE and LIMIT_SIZE?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/minicluster/DriverUtils.java
```diff
@@ -137,12 +123,8 @@ public static void dumpGraph(JobConf conf, String inputGraph, String outputFasta
         LOG.info("Dump graph to fasta took " + GenomixJobConf.tock("dumpGraph") + "ms");
     }
 
-    public static int getNumCoresPerMachine(GenomixJobConf conf) {
-        return conf.get(GenomixJobConf.HYRACKS_IO_DIRS).split(",").length;
-    }
-
     public static String[] getSlaveList(GenomixJobConf conf) {
-        return conf.get(GenomixJobConf.HYRACKS_SLAVES).split("\r?\n|\r"); // split on newlines
+        return conf.get(GenomixJobConf.SLAVES).split("\r?\n|\r"); // split on newlines
```
anbangx, line 127: This expression means: 1. \r; 2. \r\n; 3. \n

It is because the new line in different OS represents different?

JavierJia:
See, there is a green check mark!

.travis.yml
jakebiesinger, line 25: Yeah good point.  Actually, I think we should have even more than this.  We can have this simple testcase, the default pattern, and some really crazy patterns just to make sure everything is working (aka, stress test!)

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/minicluster/DriverUtils.java
jakebiesinger, line 71: They are set in the `conf/cluster.properties` file, which is then set in this for loop.

I could also explicitly set all of them instead of using the for loop.  Not sure that's actually better though.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/minicluster/DriverUtils.java
jakebiesinger, line 127: yep.  The regex matches either `\r\n`, `\n` and `\r` which are the 3 different valid styles of newlines.

jakebiesinger:
Dobule-green!  Both JDK's are working!  Woohoo!

.travis.yml
jakebiesinger, line 25: Done. Added default pipeline and an extended pipeline.

jakebiesinger:
We good to merge this in?

jakebiesinger:
Merging now, since no one else had comments... The (new) failure is some error unrelated to these commits (it's a more extended pipeline).

PR #49, opened 2013-11-01
Anbangx/debug split repeat2

anbangx:
Eventually, fix the error showed in split repeat. Actually the error is not in split repeat, but it's in path merge. The graph becomes asymmetric after path merge..... Spend a lot time figuring out -- remote debug, symmetric checking, extract subgraph, logging.... And huge non-human-readable graph... Anyway, this bug is fixed. 

Some changes in this branch:
1. Make SymmetryCheck more useful. It will not only check if the corresponding edge exists, but also check if the EdgeMap is the same one(same EdgeType and same readIdSet).
1. Add log to Split Repeat.

@jakebiesinger @JavierJia @Nan-Zhang Have a look! Thanks.


genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/checker/SymmetryCheckerVertex.java
```diff
@@ -31,28 +34,53 @@ public void initVertex() {
         outFlag = 0;
     }
 
+    public void sendEdgeMap(DIR direction) {
+        VertexValueWritable vertex = getVertexValue();
+        for (EDGETYPE et : direction.edgeTypes()) {
+            for (VKmer dest : vertex.getEdgeMap(et).keySet()) {
+                //                outgoingMsg.reset();
```
jakebiesinger: no msg.reset()?  even if there are no fields that aren't overwritten below, it's clearer to the reader if you include a reset, IMO.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/io/VertexValueWritable.java
```diff
@@ -156,16 +156,16 @@ public void readFields(DataInput in) throws IOException {
...
-            boolean verbose = false;
-            for (VKmer problemKmer : problemKmers) {
-                verbose |= this.getInternalKmer().equals(problemKmer);
-                verbose |= findEdge(problemKmer) != null;
-            }
-            if (verbose) {
-//                LOG.fine("VertexValue.readFields: " + toString());
-            }
-        }
+//        if (DEBUG) {
```
jakebiesinger: We should decide on a final version for this debugging stuff.  We can make lots of changes in our own branches but what goes into the main branch should be the "right" way of doing it.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/DeBruijnGraphCleanVertex.java
```diff
@@ -78,6 +78,8 @@ public void initVertex() {
     }
 
     public void checkDebug() {
+        if(getVertexId().toString().equals("AGACCCATGCAGCGGGTCAGG"))
```
jakebiesinger: please don't commit this kind of stuff (feel free to use it in your machine, but be selective about what gets committed!)

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/splitrepeat/SplitRepeatVertex.java
```diff
@@ -251,13 +283,29 @@ public void responseToRepeat(Iterator<SplitRepeatMessage> msgIterator) {
...
+            }
         }
     }
 
     @Override
     public void compute(Iterator<SplitRepeatMessage> msgIterator) {
         initVertex();
+        if (verbose)
+            LOG.fine("Iteration " + getSuperstep() + " for key " + getVertexId());
+        if(getVertexId().toString().equals("AGACCCATGCAGCGGGTCAGG")
```
jakebiesinger: and this stuff shouldn't be here.

jakebiesinger:
So where is the line that fixes the pathmerge bug?  You should submit it as a separate request or at least point it out somehow.  This pull request seems to be just adding logging and the symmetry checker (and several of the logging things shouldn't be there).

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/splitrepeat/SplitRepeatVertex.java
jakebiesinger: Removed.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/DeBruijnGraphCleanVertex.java
jakebiesinger: Removed.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/io/VertexValueWritable.java
jakebiesinger: Added this back in.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/checker/SymmetryCheckerVertex.java
jakebiesinger: I've added this back in.

jakebiesinger:
Merging this in.

PR #51, opened 2013-11-04
Anbangx/reduce message communicate

anbangx:
@Nan-Zhang @JavierJia @jakebiesinger 

In order to reduce the overhead of serialization and deserialization from Message communication, I add a byte flag to only read and write the fields we need. Please have a look at it. Thanks!

@JavierJia Please reprofile the pathmerge from this branch. Let's check if it has some performance optimization.

BTW, when I send pull request, can I only compare with a particular commit instead of our main branch?

PR #53, opened 2013-11-05
Fix PathMerge problem

jakebiesinger:
@anbangx @JavierJia @Nan-Zhang @Elmira88 

This PR is to fix the regression introduced by #51 commit 447146c. The stupid commit I've added here isn't the right way to fix the bug but it works for now.  I'm adding this PR so people can use this commit to directly test things.

When you do your upcoming testing, everyone, base it on this branch, please.

@anbangx I feel like it would be easier for us to keep track of these values if instead of a `byte` member variable, we just had the missing values initialized as `null`.  That involves more allocation and therefore GC and might make some pieces harder to deal with (you can't directly get contained values in e.g., `outgoingMsg`) but it would sure make reading/writing easier/  You'd just have to check if the value is `null` rather than actually read anything.

I've noticed in recent builds that I get a malformed byte error from HashMapWritable-- I bet there is a similar (or same?) issue there.

### DIFF
--- a/genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/pathmerge/BasicPathMergeVertex.java
+++ b/genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/pathmerge/BasicPathMergeVertex.java
@@ -10,6 +10,7 @@ import org.apache.commons.lang3.StringUtils;
 import edu.uci.ics.genomix.pregelix.io.VertexValueWritable;
 import edu.uci.ics.genomix.pregelix.io.VertexValueWritable.State;
 import edu.uci.ics.genomix.pregelix.io.message.PathMergeMessage;
+import edu.uci.ics.genomix.pregelix.io.message.PathMergeMessage.PATHMERGE_MESSAGE_FIELDS;
 import edu.uci.ics.genomix.pregelix.log.LogUtil;
 import edu.uci.ics.genomix.pregelix.operator.DeBruijnGraphCleanVertex;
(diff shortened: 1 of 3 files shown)


genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/pathmerge/BasicPathMergeVertex.java
```diff
@@ -135,6 +136,7 @@ public void updateNeighbors() {
                         EdgeMap edgeMap = new EdgeMap();
                         edgeMap.put(iter.next(), vertex.getEdgeMap(updateEdge).get(dest));
                         outgoingMsg.getNode().setEdgeMap(newEdgetype, edgeMap); // copy into outgoingMsg
```
anbangx, line 138: How about adding outgoingMsg.setEdgeMap(newEdgetype, edgeMap) inside MessageWriable and changing bit inside message?

jakebiesinger, line 138: The problem is, are you really going to remember that every time you change something on `node`?  There are other values in `node` (not just edgemap) that could be changed.  Why have a getNode() at all if you can't actually use the `Node` that gets returned without special casing whether or not it's saved?

Seems like it's better to use the getter/setters and keep track of what's been accessed/changed that way.

jakebiesinger, line 138: Implemented in #56 which supersedes this PR.

PR #55, opened 2013-11-06
Simplify travis scripts; fix -runLocal for driver; use -localOutput as home for STATS, DUMP_FASTA, and PLOT_SUBGRAPH

jakebiesinger:
@anbangx @Nan-Zhang @JavierJia 

Sorry to submit these three separate ideas in one PR.
- With these changes, the `-runLocal` option should no longer stall when running the driver.  I still don't have a solution for test cases run within eclipse, where I occasionally see the tests hang.  We shouldn't see the "no output received in 10 minutes" error you see in travis right now.
- When we separated the unit tests from the pipeline tests, we lost the actual command that was failing for pipeline tests.  Travis now prints it.
- `STATS`, `DUMP_FASTA`, and `PLOT_SUBGRAPH` now all write to HDFS but will be copied to the local output directory if the user specifies `-localOutput`.  Their naming is also simplified to be clearer (if the last clean step had output `05-MERGE`, then these steps will be outputted as `05-MERGE-DUMP_FASTA` rather than `06-DUMP_FASTA`. That is, the basename is kept for these jobs that don't procude their own output. This should  help our testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ install:
     - cat ~/.ssh/id_rsa.pub >> ~/.ssh/authorized_keys  # no pw required for ssh'ing locally
     - ssh localhost -o StrictHostKeyChecking=false -n  # allow us to ssh later w/out typing "yes"
     - sudo apt-get install graphviz
-    - mvn install -q -am -pl genomix/genomix-driver/ -DskipTests=true # only install driver's prereq's
+    - mvn install -am -pl genomix/genomix-driver/ -DskipTests=true # only install driver's prereq's
 
 before_script:
     - cd genomix/genomix-pregelix/
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,9 @@ jdk:
     - oraclejdk7
     - openjdk7
 env:
-    - UNIT_TESTS=true PIPELINE_TESTS=false RUN_LOCAL=""
-    - UNIT_TESTS=false PIPELINE_TESTS=true RUN_LOCAL=""
-    - UNIT_TESTS=false PIPELINE_TESTS=true RUN_LOCAL="-runLocal"
+    - UNIT_TESTS=true
+    - PIPELINE_TESTS=true
+    - PIPELINE_TESTS=true RUN_LOCAL="-runLocal"
 
 # run steps
 install:
--- a/genomix/genomix-hadoop/src/main/java/edu/uci/ics/genomix/hadoop/converttofasta/ConvertToFasta.java
+++ b/genomix/genomix-hadoop/src/main/java/edu/uci/ics/genomix/hadoop/converttofasta/ConvertToFasta.java
@@ -1,20 +1,26 @@
 package edu.uci.ics.genomix.hadoop.converttofasta;
 
+import java.io.DataOutputStream;
 import java.io.IOException;
 import java.util.logging.Logger;
 
+import org.apache.hadoop.fs.FSDataOutputStream;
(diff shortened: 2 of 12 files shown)


jakebiesinger:
No comments on these?

JavierJia:
I'm preparing the midterm in these two days. I rely on @anbangx and @Nan-Zhang  to check these changes. 
I'm coming back this evening :back:

jakebiesinger:
@anbangx I know you've been looking at these changes even before they were a pull request.  Can you get back to me quickly about them?

genomix/genomix-hadoop/src/main/java/edu/uci/ics/genomix/hadoop/converttofasta/ConvertToFasta.java
```diff
@@ -28,13 +36,58 @@
...
+
+            public synchronized void close(Reporter reporter) throws IOException {
+                out.close();
+            }
+        }
+
+        public RecordWriter<K, V> getRecordWriter(FileSystem ignored, JobConf job, String name, Progressable progress)
+                throws IOException {
+            Path file = new Path(FileOutputFormat.getTaskOutputPath(job, name), OUTPUT_FILENAME);
+            FileSystem fs = file.getFileSystem(job);
```
anbangx, line 83: What's the difference between FileSystem dfs = FileSystem.get(conf) and these two line?

jakebiesinger, line 83: I think they're doing the same thing.  I'm not sure which is more appropriate-- these two lines are based on a snippet found online. Maybe @Nan-Zhang, the resident expert on hadoop, knows?

.travis.yml
```diff
@@ -19,7 +19,7 @@ install:
     - cat ~/.ssh/id_rsa.pub >> ~/.ssh/authorized_keys  # no pw required for ssh'ing locally
     - ssh localhost -o StrictHostKeyChecking=false -n  # allow us to ssh later w/out typing "yes"
     - sudo apt-get install graphviz
-    - mvn install -q -am -pl genomix/genomix-driver/ -DskipTests=true # only install driver's prereq's
```
anbangx, line 22: Why do you delete -q? Not only show error message?

jakebiesinger, line 22: Yeah, like the relevant commit says, travis was killing our jobs for not having anything on stdout during the mvn build process.  Now the travis logs are longer but the jobs won't be killed for timeouts that are really just long builds.

genomix/genomix-hadoop/src/main/java/edu/uci/ics/genomix/hadoop/converttofasta/ConvertToFasta.java
anbangx, line 83: Answer me this question. Others look good to me!

jakebiesinger, line 83: This version returns the particular `FileSystem` that the job's output path belongs to.  I'm not aware of this ever being different than HDFS but I guess it could be?

In _this_ case where I have a RecordWriter that's acting as output format for a job, it seems to be more appropriate to get whatever the job is configured for, rather than the `FileSystem.get(conf)` which will be explicitly whatever filesystem the conf is set up for.

jakebiesinger:
Okay I'm going to go ahead and merge this in.

PR #56, opened 2013-11-07
In pregelix messages, use null values to indicate usage

jakebiesinger:

Rather than using a byte to store what fields are in use in pregelix messages, this PR sets them all to `null` and transparently populates them when a `get()` is called.  This will work as long as any reused mesages are `.reset()`'ed.

This way of storing values is also safer since in some cases, trying to read from a value that has never been set will result in a NPE (only works for values that aren't automatically generated, such as primitive types and enums). This allowed me to find and fix 2 bugs in our pregelix code.

We could take this one step further and remove the automatic creation of the msg values.

This PR supersedes #53.


jakebiesinger:
So no comments from anyone?

anbangx:
LGTM!

JavierJia:
:+1:

Nan-Zhang:
LGTM

PR #58, opened 2013-11-12
remove garbage GAGETEST.fasta

jakebiesinger:
@Nan-Zhang


Nan-Zhang:
LGTM. I think this file can be removed, cause it will created automatically when testing gage statistics

PR #64, opened 2013-11-16
fix fasta output to use internal kmer not key

jakebiesinger:
@Nan-Zhang


jakebiesinger:
@anbangx do you want to okay this?  @Nan-Zhang seems absent :)

anbangx:
LGTM

PR #65, opened 2013-11-16
Fix for "missing jar" problem

jakebiesinger:
@anbangx 
We started cleaning up the `HDFS_WORK_PATH` in #62 but were doing do _after_ deploying our jars there.


anbangx:
LGTM

PR #67, opened 2013-11-20
increase hardcoded pregelix framesize (tmp fix)

jakebiesinger:
Hi all, if you're planning on running our pipeline on decently sized data and will be doing a lot of merging, you're going to need to make adjustments to the pregelix's hard-coded bufferCache pageSize.  Yingyi is working on a way to configure this, but for now just make a change similar to what  you see in the diff. 

This will solve problems that look like:

``` java
Caused by: edu.uci.ics.hyracks.storage.am.common.api.TreeIndexException: Space required for record (137070) larger than maximum acceptable size (131059)
```

### DIFF
--- a/genomix/genomix-driver/src/main/resources/conf/cluster.properties
+++ b/genomix/genomix-driver/src/main/resources/conf/cluster.properties
@@ -31,7 +31,7 @@ CLASSPATH="${HADOOP_HOME}:${CLASSPATH}:."
 # The FRAME_SIZE is the size in bytes of a single "page" used in an external merge-sort.
 # The FRAME_LIMIT is the number of such pages to use in an in-memory sort buffer.
 # By default, we use 64k * 4096 = 268MB buffer.
-FRAME_SIZE=65535
+FRAME_SIZE=1048560
 FRAME_LIMIT=4096
 
 # The number of jobs to keep logs for (logs are kept in-memory)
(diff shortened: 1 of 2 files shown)


JavierJia:
LGTM. Do we have some way to remind us to change it back when we have the better solution ?

jakebiesinger:
@JavierJia when it's fixed and we merge with hyracks `master`, there will be a conflict from this commit and we'll just accept whatever incoming changes @sigmod has made.

I've also pushed a change that increases the default FRAME_LIMIT.  It's pretty big (1MB), making the sort buffer huge (1MB \* 4096 = 4G) but will save us some headaches later.  What do you think?

JavierJia:
For the Frame_limit, I think it don't have to be very big. The 3 frame for merge sort is enough. The FRAME_LIMIT will define how many frame it will compare at the same time. If we want to make the merge sort finished at one round, we just need to control : FrameSize >= FrameNum \* sizeof(one tuple).

jakebiesinger:
Yes, you are right-- my commit message is wrong.  It should read `increase default FRAME_SIZE` not `FRAME_LIMIT`.  I don't touch FRAME_LIMIT, though as you say it could be smaller than 4k at the cost of taking more iterations in the merge sort.

JavierJia:
Then LGTM :+1:

jakebiesinger:
Open 4 days and no other comments.  I'll go ahead and merge this.

PR #70, opened 2013-11-23
update KMP and change name to indexOf()

Nan-Zhang:
@jakebiesinger @JavierJia @anbangx @Elmira88  separated KMP code and create a new branch for it.

### DIFF
--- a/genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/VKmer.java
+++ b/genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/VKmer.java
@@ -538,7 +538,7 @@ public class VKmer extends BinaryComparable implements Serializable, WritableCom
         revcomp.setReversedFromStringBytes(kmer.getKmerLetterLength(), kmer.toString().getBytes(), 0);
         mergeWithFFKmer(initialKmerSize, revcomp);
     }
-    
+
     /**
      * Merge Kmer with the next connected Kmer, when that Kmer needs to be
      * reverse-complemented e.g. AAGCTAA merge with GGTTGTT, if the initial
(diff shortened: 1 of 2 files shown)


jakebiesinger:
It would be nice to have longer sequences in your test cases.  The sequences we'll be running this against will be long (100-250bp) as will the kmers (21-55bp).

Nan-Zhang:
@jakebiesinger  already added longer read test for IndexOf(). Can this branch be merger ?

jakebiesinger:
Yep, LGTM.  What about @JavierJia's question?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/VKmer.java
```diff
@@ -727,6 +727,64 @@ public int editDistance(VKmer other) {
         return editDistance(this, other);
     }
 
+    public int indexOf(VKmer pattern) {
+        return indexOf(this, pattern);
```
anbangx, line 731: I think it is better to move this indexOf(text, pattern) out of VKmer and make a new class for it, maybe StringMatch... When you try to use it in our GA, pass parameters. Since you have made KMP algorithm to static, it is the same to make a new container. But a new class is clear.

genomix/genomix-data/src/test/java/edu/uci/ics/genomix/type/VKmerTest.java
```diff
@@ -609,5 +610,67 @@ public void TestLargeKmerMergeRR() {
...
+        Random random = new Random();
+        char[] buf = new char[length];
+        for (int idx = 0; idx < buf.length; idx++) {
+            buf[idx] = symbols[random.nextInt(4)];
+        }
+        return new String(buf);
+    }
+    
+    @Test
+    public void TestIndexOfForShortRead(){
```
anbangx, line 632: If you make StringMatch as a new class, you can test here by using text and pattern. It will much more readable.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/VKmer.java
JavierJia, line 731: I think it will be fine here. Because we don't want to parse the kmer into string and then compare each char. we can directly get gene character in Kmer.

JavierJia:
LGTM. Please take care of Anbang's comments.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/VKmer.java
jakebiesinger, line 731: I agree.  It's a VKmer search we're doing, not a generic one.

Nan-Zhang, line 731: directly manipulating byte code is much faster than manipulating string.

anbangx, line 731: OK. Make sense.

anbangx:
LGTM

PR #75, opened 2013-11-29
Latest merge with master plus pregelix API changes

jakebiesinger:
I made some changes to the pregelix API which have now been merged in to Yingyi's branch so this PR brings in the latest hyracks master along with usability changes:
- pregelix now has an API for specifying all combinations of `wake`/`sleep` and `update-state`/`use-previous-state`.  These are accessible via the new `activate(boolean update)` and `voteToHalt(boolean update)` interfaces.  There were cases (especially in search) where the vertex should go to sleep and no changes were made to its value.  This API should help us be more explicit and efficient.
- fixed bug in pregelix where custom aggregators weren't being used if specified in xml via `addResources` (this affected our pregelix test cases).
- add a simple hadoop Counters-based aggregator which aggregates across all iterations of a job. We specify conversion from a Vertex<> to a Counter and add a call to `job.setCounterAggregatorClass(SomeCounterAgg.class)` and pregelix will now do the rest.

Once these changes are in place, we can update our `StatisticsAggregator` and `STATS` jobs.


JavierJia:
LGTM

Nan-Zhang:
LGTM

PR #77, opened 2013-12-02
Anbangx/implement bubble merge with search

anbangx:
Hi,

This is the new version of Bubble Merge - Bubble Merge With Search.
- Ex.      B - C - D - E - F - G  
-       A - B - H - I ----------- G
- From this example, CDEF and HI are formed bubble and B and G are two end points of this bubble.
- For this operator,
-     1. it starts BFSearching from start point B, and then it will collect all the valid paths(BCDEFG and BHIG) to B
-     2. compare all the valid paths with its own startHeads and pick one path which is enough similar so that we have confidence to say this path is the correct path and others are noice
-     3. keep this similar path and merge the path nodes in B
-     4. delete all path nodes and the corresponding edges with these path nodes 

This operator needs two more FIXMES: 
1. need to add statistical method to get top 10%(or some reasonable number) startHeads as begin points. This requires to use StatisticCounter. We still wait for Yingyi's response for our StatisticCounter. Right, @jakebiesinger ?
2. how to compare the similarity.

@jakebiesinger  Have a look at it. Once the StatisticCounter can be used properly, two more FIXMEs will be finished soon.


genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
```diff
@@ -239,6 +242,7 @@ public static void verifyPatterns(Patterns[] patterns) {
     // Graph cleaning   
     public static final String BRIDGE_REMOVE_MAX_LENGTH = "genomix.bridgeRemove.maxLength";
     public static final String BUBBLE_MERGE_MAX_DISSIMILARITY = "genomix.bubbleMerge.maxDissimilarity";
+    public static final String BUBBLE_MERGE_WITH_SEARCH_MAX_LENGTH = "genomix.bubbleMergewithsearch.maxSearchLength";
```
jakebiesinger: should be `bubbleMergeWithSearch`

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
```diff
@@ -362,6 +366,9 @@ private void fillMissingDefaults() {
         if (getFloat(BUBBLE_MERGE_MAX_DISSIMILARITY, -1) == -1)
             setFloat(BUBBLE_MERGE_MAX_DISSIMILARITY, .5f);
 
+        if (getInt(BUBBLE_MERGE_WITH_SEARCH_MAX_LENGTH, -1) == -1)
+            setInt(BUBBLE_MERGE_WITH_SEARCH_MAX_LENGTH, 100);
```
jakebiesinger: Let's make the default value related to `K`.  Like `K * 2`.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/minicluster/GenerateGraphViz.java
```diff
@@ -61,7 +61,7 @@ public static GRAPH_TYPE getFromInt(int n) {
 
     public static void writeLocalBinToLocalSvg(String srcDir, String destDir, GRAPH_TYPE graphType) throws Exception {
         byte[] img = convertGraphToImg(new JobConf(), srcDir, destDir, graphType, "svg");
-        File out = new File(destDir + File.separator + "graphviz.svg");
+        File out = new File(destDir); //  + File.separator + "graphviz.svg"
```
jakebiesinger: Won't this change break the driver code, which expects `destDir` to be a directory?

genomix/genomix-pregelix/data/TestSet/BUBBLE_WITH_SEARCH/bubblemerge_withsearch/data
```diff
@@ -0,0 +1,8 @@
+TAA	{FF:{AAC=[2]}	FR:null	RF:null	RR:null	5':[2-0_0], ~5':null	kmer:null	cov:1.0x}
```
jakebiesinger: :confused: is this an input file or an expected output?  if input, where is the corresponding `bin` directory?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/io/VertexValueWritable.java
```diff
@@ -20,17 +20,26 @@
...
-        public static final byte KEEP_NODE = 0b1 << 6; 
+        public static final byte KEEP_NODE = 0b1 << 6;
         // DEAD_NODE is used in RemoveLowcoverage, if the vertex is deleted, marking it as DEAD_NODE
-        public static final byte DEAD_NODE = 0b1 << 6; 
-        
+        public static final byte DEAD_NODE = 0b1 << 6;
+
         public static final byte VERTEX_MASK = 0b1 << 6;
+
+        // UPDATE_PATH_TO_NEXT, END_NOTICE_TO_SRC and UPDATE_BRANCH_TO_SRC are used in BubbleMergeWithSearch
```
jakebiesinger: If these are only used in `BubbleMergeWithSearch`, why don't you create a subclass for them, like all the other VertexValues do?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/io/message/BubbleMergeWithSearchMessage.java
```diff
@@ -0,0 +1,163 @@
...
+        super.reset();
+        preKmerLength = null;
+        internalKmer = null;
+        pathList = null;
+        numBranches = null;
+        edgeTypeList = null;
+    }
+    
+    public Integer getPreKmerLength() {
+        if(preKmerLength == null)
```
jakebiesinger: Unless you have reason to do otherwise, you should just return an `int` not an `Integer` and you shouldn't set the value to `0` by default.  That way, if they call `getPreKmerLength()` before initializing, they'll get a NPE (which is better than the corresponding bug!)

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/io/message/BubbleMergeWithSearchMessage.java
```diff
@@ -0,0 +1,163 @@
...
+            preKmerLength = new Integer(0);
+        return preKmerLength;
+    }
+
+    public void setPreKmerLength(Integer preKmerLength) {
```
jakebiesinger: same here-- have the argument an `int` since then if they pass `null`, it will be an error.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/io/message/BubbleMergeWithSearchMessage.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+    
+    public EdgeTypeList getEdgeTypeList() {
+        if(edgeTypeList == null)
+            edgeTypeList = new EdgeTypeList();
+        return edgeTypeList;
+    }
+
+    public void setEdgeTypeList(EdgeTypeList edgeTypeList) {
+        if(edgeTypeList != null){
```
jakebiesinger: reformat

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/io/message/BubbleMergeWithSearchMessage.java
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+
+    public Integer getNumBranches() {
+        if(numBranches == null)
+            numBranches = new Integer(0);
+        return numBranches;
+    }
+
+    public void setNumBranches(Integer numBranches) {
```
jakebiesinger: `int` as arg

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/io/message/BubbleMergeWithSearchMessage.java
```diff
@@ -0,0 +1,163 @@
...
+        this.numBranches = new Integer(numBranches);
```
jakebiesinger: unboxing will happen automatically.  No need to call `new` here. Just do an assignment (with the rhs being an `int`).

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/io/message/BubbleMergeWithSearchMessage.java
```diff
@@ -0,0 +1,163 @@
...
+        }
+        if ((messageFields & BUBBLEMERGE_WITH_SEARCH_FIELDS.NUM_BRANCHES) != 0) {
+            setNumBranches(in.readInt());
+        }
+    }
+
+    @Override
+    public void write(DataOutput out) throws IOException {
+        super.write(out);
+        if (preKmerLength != null) {
```
jakebiesinger, line 121: for `write` and `getActiveMessageFields`, you can check if the lists are `null` OR if their length is `0`.  If either of those cases, don't write/read them.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,269 @@
...
+ *     have confidence to say this path is the correct path and others are noice
+ *     3. keep this similar path and merge the path nodes in B
+ *     4. delete all path nodes and the corresponding edges with these path nodes 
+ */
+public class BubbleMergeWithSearchVertex extends
+        DeBruijnGraphCleanVertex<BubbleMergeWithSearchVertexValueWritable, BubbleMergeWithSearchMessage> {
+
+    //    private static final Logger LOG = Logger.getLogger(BubbleMergeWithSearchVertex.class.getName());
+
+    private Integer MAX_BFS_LENGTH = -1;
```
jakebiesinger: if you make this `static`, you only have to call `initVertex` on the first iteration.

jakebiesinger: And it should really be an `int` not an `Integer`.  If you really want `Integer`, you should initialize it to `null`, not `-1`.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,269 @@
...
+
+/**
+ * Graph clean pattern: Bubble Merge With Search
+ * Ex.      B - C - D - E - F - G  
+ *      A - B - H ------------- G
+ * From this example, CDEF and H are formed bubble and B and G are two end points of this bubble.
+ * For this operator,
+ *     1. it starts BFSearching from start point B, and then it will collect all the valid paths(BCDEFG and BHG) to B
+ *     2. compare all the valid paths with its own startHeads and pick one path which is enough similar so that we 
+ *     have confidence to say this path is the correct path and others are noice
```
jakebiesinger: > `noice`

??

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,269 @@
...
+            }
+        }
+    }
+
+    public void updatePathInNextNode(BubbleMergeWithSearchMessage incomingMsg) {
+        BubbleMergeWithSearchVertexValueWritable vertex = getVertexValue();
+        int internalKmerLength = vertex.getInternalKmer().getKmerLetterLength();
+        VKmer source = incomingMsg.getPathList().getPosition(0);
+        int newLength = internalKmerLength + incomingMsg.getPreKmerLength() - kmerSize + 1;
+        if (newLength > MAX_BFS_LENGTH) {
```
jakebiesinger, line 153: You need some more comments on what these two branches are doing and why.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,269 @@
...
+            outgoingMsg.setPathList(incomingMsg.getPathList());
+            outgoingMsg.setInternalKmer(incomingMsg.getInternalKmer());
+            outgoingMsg.setEdgeTypeList(incomingMsg.getEdgeTypeList());
+            sendMsg(source, outgoingMsg);
+        } else {
+            outgoingMsg.reset();
+
+            // update pathList
+            VKmerList pathList = incomingMsg.getPathList();
+            pathList.append(getVertexId());
```
jakebiesinger, line 183: doesn't this mutate the `incomingMsg`? Shouldn't we make the `outgoingMsg` use a copy and then append there?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,269 @@
...
+            outgoingMsg.setPathList(pathList);
+
+            // update internalKmer
+            EdgeTypeList edgeTypes = incomingMsg.getEdgeTypeList();
+            int size = edgeTypes.size();
+            VKmer internalKmer = incomingMsg.getInternalKmer();
```
jakebiesinger: again here with the incomingMsg.  This is okay if you're _sure_ you're not going to reuse the incomingMsg anywhere but maybe it's safer to use the outgoingMsg as a copy since you have to do that anyway?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,269 @@
...
+        int internalKmerLength = vertex.getInternalKmer().getKmerLetterLength();
+        VKmer source = incomingMsg.getPathList().getPosition(0);
+        int newLength = internalKmerLength + incomingMsg.getPreKmerLength() - kmerSize + 1;
+        if (newLength > MAX_BFS_LENGTH) {
+            // send back to source vertex (pathList, internalKmer and setEdgeTypeList)
+            outgoingMsg.reset();
+            outgoingMsg.setPathList(incomingMsg.getPathList());
+            outgoingMsg.setInternalKmer(incomingMsg.getInternalKmer());
+            outgoingMsg.setEdgeTypeList(incomingMsg.getEdgeTypeList());
+            sendMsg(source, outgoingMsg);
```
jakebiesinger, line 175: Seems like this branch should also mark `END_NOTICE_IN_SRC`, right?  You need to notify `source` that this path has come to an end...

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,269 @@
...
+            if (vertex.outDegree() == 0) {
+                outgoingMsg.setEdgeTypeList(incomingMsg.getEdgeTypeList());
+                outgoingMsg.setFlag(State.END_NOTICE_IN_SRC);
+                sendMsg(source, outgoingMsg);
+                return;
+            }
+            // if numBranches > 1, update numBranches
+            if (vertex.outDegree() > 1) {
+                outgoingMsg.setFlag(State.UPDATE_BRANCH_IN_SRC);
+                outgoingMsg.setNumBranches(vertex.outDegree());
```
jakebiesinger, line 206: was the number of branches equal to 1 before?  The logic on that should be clearer, I think.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,269 @@
...
+                sendMsg(source, outgoingMsg);
+            }
+
+            // send to next 
+            EDGETYPE preToMe = incomingMsg.getEdgeTypeList().get(incomingMsg.getEdgeTypeList().size() - 1);
+            for (EDGETYPE et : preToMe.mirror().dir().mirror().edgeTypes()) {
```
jakebiesinger: once TipRemoveWithSearch is merged, `preToMe.mirror().dir().mirror().edgeTypes()` can be simplified to `preToMe.neightborDir().edgeTypes()`.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,269 @@
...
+                for (VKmer dest : vertex.getEdgeMap(et).keySet()) {
+                    // set flag and source vertex
+                    outFlag |= State.UPDATE_PATH_IN_NEXT;
```
jakebiesinger: this is the first time I've seen `outFlag`.  Why not init to `0` so that's more obvious? (plus, then you don't have to clear only certain bits here).

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,269 @@
...
+                    outFlag &= EDGETYPE.CLEAR;
+                    outFlag |= et.mirror().get();
+                    outgoingMsg.setFlag(outFlag);
```
jakebiesinger: I think it's okay to put all this `outFlag` stuff on one line like `setFlag(State.UPDATE_PATH_IN_NEXT | et.mirror().get())`.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,269 @@
...
+                    outgoingMsg.setSourceVertexId(getVertexId());
```
jakebiesinger: Isn't the `sourceVertexId` supposed to be the originator of the search?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,269 @@
...
+
+                    // update edgeTypes
+                    EdgeTypeList tmp = (EdgeTypeList) edgeTypes.clone();
+                    tmp.add(et);
+                    outgoingMsg.setEdgeTypeList(tmp);
```
jakebiesinger: why not just set it in the outgoingMsg and then `outgoingMsg.getEdgeTypList().add(et)` ?  Then you don't need `tmp` at all.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,269 @@
...
+        // update internalKmer
+        vertex.getArrayOfInternalKmer().add(incomingMsg.getInternalKmer());
+
+        // update edgeTypeList
+        vertex.getArrayOfEdgeTypes().add(incomingMsg.getEdgeTypeList());
+
+        // update numBranches
+        int numBranches = vertex.getNumBranches();
+        numBranches--;
+        vertex.setNumBranches(numBranches);
```
jakebiesinger: Why not just `vertex.getNumBranches()--`?

jakebiesinger: Maybe it's not as clear?

``` java
int numBranches = --vertex.getNumBranches();
```

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,269 @@
...
+        if (numBranches == 0) {
```
jakebiesinger, line 241: I think this branch should live in a separate function.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,269 @@
...
+            /* process in src */
+            // step1: figure out which path to keep
+            int k = 1; // FIXME compare similarity with startHead and get the most possible path, here for test, using 1
+            VKmerList pathList = vertex.getArrayOfPathList().get(k);
+            VKmer mergeKmer = vertex.getArrayOfInternalKmer().get(k);
+            EdgeTypeList edgeTypes = vertex.getArrayOfEdgeTypes().get(k);
+
+            // step2: replace internalKmer with mergeKmer and clear edges towards path nodes
+            vertex.setInternalKmer(mergeKmer);
```
jakebiesinger: I don't think it would be a replacement, right?  the nodes you've searched should be **merged** with this `internalKmer`, right?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,269 @@
...
+            vertex.getEdgeMap(vertex.getArrayOfEdgeTypes().get(k).get(0)).remove(pathList.getPosition(1));
```
jakebiesinger: Also, this shouldn't be just a removal of that edge.  It should be a replacement with the end of this branches' edges.

```
   ----B<E
A*---C<F
  ---D<G
```

If the search terminates at B, C, and D and you choose to keep B, you should end up with `A<E` right?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,269 @@
...
+
+            // step3: send kill message to path nodes
```
jakebiesinger: I think we're a bit confused on the right logic here.  The bubble should only merge those nodes that are similar enough.  The non-similar set shouldn't be removed (and may not even be detached from this path!)

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,269 @@
...
+            for (int i = 1; i < pathList.size(); i++) {
+                VKmer dest = new VKmer(pathList.getPosition(i));
+
+                outgoingMsg.reset();
+                outgoingMsg.setFlag(State.KILL_MESSAGE_FROM_SOURCE);
+
+                // prev stores in pathList(0)
+                VKmerList kmerList = new VKmerList();
```
jakebiesinger: could be `clear`ing instead of reallocating.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,269 @@
...
+        BubbleMergeWithSearchVertexValueWritable vertex = getVertexValue();
+
+        // send msg to delete edges except the path nodes
+        for (EDGETYPE et : EDGETYPE.values()) {
+            for (VKmer dest : vertex.getEdgeMap(et).keySet()) {
+                if ((et == incomingMsg.getEdgeTypeList().get(0) && dest
+                        .equals(incomingMsg.getPathList().getPosition(0)))
+                        || (incomingMsg.getEdgeTypeList().size() == 2 && et == incomingMsg.getEdgeTypeList().get(1) && dest
+                                .equals(incomingMsg.getPathList().getPosition(1))))
+                    continue;
```
jakebiesinger, line 312: I have no idea what case you're skipping here.  Some comments would be nice!

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,269 @@
...
+                outgoingMsg.reset();
+                outFlag |= State.PRUNE_DEAD_EDGE;
+                outFlag |= et.mirror().get();
+                outgoingMsg.setFlag(outFlag);
+                outgoingMsg.setSourceVertexId(getVertexId());
+                sendMsg(dest, outgoingMsg);
```
jakebiesinger, line 318: Have you done any testing on what happens if a node that's deleted in one iteration receives a message in the next iteration?  (I hope it doesn't wake up!)

genomix/genomix-pregelix/src/test/java/edu/uci/ics/genomix/pregelix/jobgen/JobGenerator.java
```diff
@@ -208,6 +211,19 @@ private static void genBubbleMergeGraph() throws IOException {
         generateBubbleMergeGraphJob("BubbleMergeGraph", outputBase + "BUBBLE.xml");
     }
 
+    private static void generateBubbleMergeWithSearchGraphJob(String jobName, String outputPath) throws IOException {
+        PregelixJob job = BubbleMergeWithSearchVertex.getConfiguredJob(new GenomixJobConf(3),
+                BubbleMergeWithSearchVertex.class);
+        job.setVertexInputFormatClass(NodeToBubbleMergeWithSearchVertexInputFormat.class);
```
jakebiesinger: All format changes should be done in `getConfiguredJob`.

jakebiesinger:
The idea/implementation seems sound. I think you need a better description of what's going on here. Maybe you could add something like this to your main `Vertex` header?

> We do a breadth-first search by passing along messages. All paths are explored, accumulating a long kmer that represents the sequence of that path. Each branch searches up to a certain kmer distance away and the source node keeps track of how many active branches there are. When there are no more active branches from this path, we compare the branches to detect similar paths and collapse those dubbed both too similar and inferior in some way (e.g., too low coverage).

jakebiesinger:
There are two problems I see beyond the stubs you mentioned in the code:
1. you don't detect intersection between different sources
2. not clear how to deal with loops in the graph (it seems you'll just walk on through them which might wreak havoc in your graph...)

For point 2, think of this example:

```
A-ff->B-fr->B-rr->A
```

When you go to delete nodes, you might end up deleting the source node!

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
anbangx: Done!

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/config/GenomixJobConf.java
anbangx: Done

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/minicluster/GenerateGraphViz.java
anbangx: Here, I just want to use "graphviz.svg" as destFile. Why do we need destDir here?

genomix/genomix-pregelix/data/TestSet/BUBBLE_WITH_SEARCH/bubblemerge_withsearch/data
anbangx: REMOVE!

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/io/VertexValueWritable.java
anbangx: Add BubbleMergeWithSearchState. Done

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/io/message/BubbleMergeWithSearchMessage.java
anbangx: I change all the Integer to int and initialize as -1

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/io/message/BubbleMergeWithSearchMessage.java
anbangx: Done

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/io/message/BubbleMergeWithSearchMessage.java
anbangx: Done

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/io/message/BubbleMergeWithSearchMessage.java
anbangx: Done

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/io/message/BubbleMergeWithSearchMessage.java
anbangx: Good point! Done!

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/io/message/BubbleMergeWithSearchMessage.java
anbangx, line 121: Done!

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
anbangx: Done!

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
anbangx, line 153: Done

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
anbangx, line 175: Yep. Add it!

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
anbangx, line 183: I never mutate incomingMsg.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
anbangx, line 206: Here, initially numBranches = 1, when there is any branch(says n), it will update numBranches to numBranches + n - 1

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
anbangx: Got it!

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
anbangx: Done

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
anbangx: Done

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
anbangx: sourceVertexId is supposed to be the originator of the search. It doesn't need to set SourceVertexId here.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
anbangx: Done

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
anbangx: Java doesn't allow to do --vertex.getNumBranches()

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
anbangx, line 241: Add the separate function

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
anbangx: OK. I change the logic to only prune other edges that connecting with similar set

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
anbangx: yep

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
anbangx: Done!

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
anbangx, line 312: Done!

genomix/genomix-pregelix/src/test/java/edu/uci/ics/genomix/pregelix/jobgen/JobGenerator.java
anbangx: getConfiguredJob() only sets in this way  
job.setVertexInputFormatClass(NodeToVertexInputFormat.
job.setVertexOutputFormatClass(VertexToNodeOutputFormat.class);

anbangx:
Some changes based on Jake's comments.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/minicluster/GenerateGraphViz.java
jakebiesinger: It's not necessary but rather the way the Driver steps are set up-- for all steps, you specify an input _dir_ and an output _dir_.  The driver calls this function directly.

I don't mind either way but did you update the other users of this code to use "graphviz.svg" instead of the destDir?

genomix/genomix-pregelix/data/TestSet/BUBBLE_WITH_SEARCH/bubblemerge_withsearch/data
jakebiesinger: So there are no `TestSet`s now for Bubble Merge?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
jakebiesinger, line 183: yes, you do right here!

``` java
VKmerList pathList = incomingMsg.getPathList();
pathList.append(...);
```

pathList is a reference to `incomingMsg`'s list, yes?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
jakebiesinger, line 206: this code isn't being run in `source` though... it's being run in one of the message frontiers.  How can it possibly be up-to-date on how many branches there are active?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
jakebiesinger: It allows `int numBranches = vertex.getNumBranches()--;`. I just did it :)

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
jakebiesinger: Did you fix this?

genomix/genomix-pregelix/src/test/java/edu/uci/ics/genomix/pregelix/jobgen/JobGenerator.java
jakebiesinger: That's why you need to create a new version of `getConfiguredJob` that handles your classes' specific vertex input/output needs, just like every other `Vertex` type.  Otherwise, the driver won't work!

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/minicluster/GenerateGraphViz.java
anbangx: Why do we need .svg? Can we keep it graphviz?

genomix/genomix-pregelix/src/test/java/edu/uci/ics/genomix/pregelix/jobgen/JobGenerator.java
anbangx: Done

anbangx:
For the loop in graph, like 
A-ff->B-fr->B-rr->A

What result do we expect?

anbangx:
1. Add comments to vertex header;
2. Add detection to detect intersection between different sources 
3. TODO: how to deal with loop case?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/minicluster/GenerateGraphViz.java
jakebiesinger: specifying the extension lets you be more flexible-- sometimes we may want `.png` or `.dot` formats rather than `.svg`.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
jakebiesinger: Seems this is still broken.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
jakebiesinger, line 318: No comment here?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,313 @@
...
+
+    public void pickBestPath() {
+        BubbleMergeWithSearchVertexValueWritable vertex = getVertexValue();
+
+        // step1: figure out which path to keep
+        int k = 1; // FIXME compare similarity with startHead and get the most possible path, here for test, using 1
+        VKmerList pathList = vertex.getArrayOfPathList().get(k);
+        EdgeTypeList edgeTypes = vertex.getArrayOfEdgeTypes().get(k);
+
+        // step2: clear edges except that one towards similar path
```
jakebiesinger: so you no longer merge with the adjacent kmer at all?  you just remove the other paths out of this node and kill the bubbles?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,313 @@
...
+
+            // store edgeType in msg.ArrayListWritable<EDGETYPE>(0)
+            outgoingMsg.getEdgeTypeList().add(edgeTypes.get(i - 1).mirror());
+            if (i < edgeTypes.size())
+                outgoingMsg.getEdgeTypeList().add(edgeTypes.get(i));
+            sendMsg(dest, outgoingMsg);
+        }
+    }
+
+    public void sendKillMsgToPathNodes(BubbleMergeWithSearchMessage incomingMsg) {
```
jakebiesinger: This is NOT a kill message, right?  It's a "remove all edges that aren't in the incomingMsg".  Maybe you can call that "saveOnlyPathEdges"

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
```diff
@@ -0,0 +1,313 @@
...
+    public void pickBestPath() {
+        BubbleMergeWithSearchVertexValueWritable vertex = getVertexValue();
+
+        // step1: figure out which path to keep
+        int k = 1; // FIXME compare similarity with startHead and get the most possible path, here for test, using 1
+        VKmerList pathList = vertex.getArrayOfPathList().get(k);
+        EdgeTypeList edgeTypes = vertex.getArrayOfEdgeTypes().get(k);
+
+        // step2: clear edges except that one towards similar path
+        for (EDGETYPE et : EDGETYPE.OUTGOING) {
```
jakebiesinger: There's no such thing as OUTGOING.  It's `FORWARD.edgeTypes()`, right

jakebiesinger: The overall direction should be a parameter.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
jakebiesinger: Post-discussion, this is clearer.  I still think you need a high-level description of **all** the steps.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
anbangx, line 183: Yep. and it does deep copy in pathList.append()

jakebiesinger, line 183: what?  You might make a copy of the VKmer but you're modifying the oncomingMsg.pathList here.  It can't do a deep copy of itself...

anbangx, line 183: Oh, yeah, I will fix it.

jakebiesinger:
I think your current action of not allowing paths from different sources to
intersect will handle this case for now-- you won't be able to go backwards
at all since you've already visited those nodes.

Can you add a `TODO` message that says "handle self- and extended-loops
more reasonably"?

## 

Jake Biesinger
Graduate Student
Xie Lab, UC Irvine

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
jakebiesinger: you also mutate the `incomingMsg.internalKmer` here.  Like I said before, you should copy it to `outgoingMsg` or keep a local _copy_, not reference if you're going to do a `mergeWithKmerInDir`, which mutates the VKmer.

jakebiesinger:
Now let's hurry on this one so I can merge the refactor in without having to redo your work here.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/bubblemerge/BubbleMergeWithSearchVertex.java
anbangx: Done

anbangx:
Some refactors based on Jake's comments!

anbangx:
Some changes:
1. Store string as command line option for SEARCH_DIRECTION
2. Add the code to deal with the case when pruning last node's edges

PR #76, opened 2013-11-29
cleanup statistics aggregators

jakebiesinger:
@anbangx @JavierJia @Elmira88 @Nan-Zhang 
All statistics can now be handled in the compute functions hadoop Counters, like so:
- `getCounters().findCounter(GraphMutations.Num_RemovedBridges).increment(1);`
- or equivalently but slower: `getCounters().findCounter("GraphMutations", "Num_RemovedBridges").increment(1);`
- or any group and counter name: `getCounters().findCounter("genomix", "coolness-factor").increment(Long.MAX_VALUE);`

PR #78, opened 2013-12-02
Anbangx/fix aggregator

anbangx:
Hi,

Some small changes in this branch:
1. add a script to automatically generate times and size from log
2. change to use only one nc and only one dev in pregel test suite. Make test easy.
3. fix aggregator - Don't use preReadAggregator anymore. But like Jake mentioned before, the aggregator in pregel is only allowed in single iteration. If we want to use cross iteration, it will be very expensive. Thank you Jake for this PR #76 . Hope it can solve this problem.


jakebiesinger:
@anbangx #76 has been merged in.  You should merge with the main branch and update your methods to use the `Counters`-based aggregators.

jakebiesinger:
@anbangx is this PR necessary anymore?  Maybe resubmit with just the changes you need?

anbangx:
Not necessary. I moved what I need to another branch.

PR #82, opened 2013-12-04
Elmira/implementing ray

Elmira88:
Hi!
@jakebiesinger @Nan-Zhang @anbangx @JavierJia 
This branch is the first try for Ray Implementation, and still in progress.

-It's a little messy. :)


jakebiesinger:
@anbangx I think we should merge this into the main branch despite it being only partially complete-- that way the reorganization can happen only once without having to re-merge things that affect this branch.

anbangx:
I agree!

On Tue, Dec 3, 2013 at 9:09 PM, Jake Biesinger notifications@github.comwrote:

> @anbangx https://github.com/anbangx I think we should merge this into
> the main branch despite it being only partially complete-- that way the
> reorganization can happen only once without having to re-merge things that
> affect this branch.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/uci-cbcl/genomix/pull/82#issuecomment-29780009
> .

## 

Best Regards,

Anbang Xu

PR #84, opened 2013-12-04
add test for union update of vkmerlist

Nan-Zhang:
@jakebiesinger


Nan-Zhang:
@jakebiesinger  test passed~

jakebiesinger:
fix the test cases as we discussed, then LGTM.

PR #85, opened 2013-12-05
Wbiesing/bugfix for p4

jakebiesinger:

I found a regression in P4-merge, likely cause by the recent refactor.  In the old code, `newEdgetype` was initialized and is where this node's neighbor should be attached to `outgoingMsg`. Instead, it was being attached to `updateEdge`.

I fixed this and restructured the loop to 1) reduce the number of inner iterations and 2) to ensure that `outgoingMsg` is being reset for each iteration, just to be safe.

Also, @Elmira88, the StatisticsAggregator is no longer used and has been removed.  I removed the import; you can remove the commented out code :)


Elmira88:
OK! Thanks.

PR #83, opened 2013-12-04
Anbangx/clean code

anbangx:
@jakebiesinger @JavierJia @Nan-Zhang @Elmira88 

This PR reorganizes our whole package, especially pregel-package. Now it looks like much more readable and the structure is much clearer.

PR #86, opened 2013-12-05
Jianfeng/node store in hdfs

JavierJia:
@jakebiesinger @anbangx @Nan-Zhang @Elmira88 

Hi team,

This PR is to change the ReadHeadSet into a ExternalableSet which can be stored in the HDFS.
I make the genomix-hyracks test passed. I will continue working on the other test cases. I'd like to merge it to master first to let us use this collection in the scaffolding.


genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/BTreeSet.java
```diff
@@ -0,0 +1,431 @@
...
+
+    private static final ThreadFactory threadFactory = new ThreadFactory() {
+        public Thread newThread(Runnable r) {
+            return new Thread(r);
+        }
+    };
+
+    protected class BTreeStorageManager {
+
+        private static final int PAGE_SIZE = 4096;
```
jakebiesinger: possible to read these values from a conf?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/BTreeSet.java
```diff
@@ -0,0 +1,431 @@
...
+        RangePredicate searchRange = new RangePredicate();
+        while (scanCursor.hasNext()) {
+            scanCursor.next();
+            searchRange.setLowKey(scanCursor.getTuple(), true);
+            searchRange.setHighKey(scanCursor.getTuple(), true);
+            searchRange.setLowKeyComparator(comparator);
+            searchRange.setHighKeyComparator(comparator);
+
+            searchCursor.reset();
+            larggerAccessor.search(searchCursor, searchRange);
```
jakebiesinger: should be `largerAccessor`

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/Node.java
```diff
@@ -373,10 +371,10 @@ public int setAsReference(byte[] data, int offset) {
             }
         }
         if ((activeFields & NODE_FIELDS.UNFLIPPED_READ_IDS) != 0) {
-            offset = getUnflippedReadIds().setAsCopy(data, offset);
+            offset += getUnflippedReadIds().setAsCopy(data, offset);
```
jakebiesinger: why `+=` here?  all `setAsCopy` and `setAsReference` methods return the new offset, not the length...

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
```diff
@@ -1,88 +1,64 @@
...
+import java.util.Set;
 import java.util.SortedSet;
-import java.util.TreeSet;
 
-import org.apache.hadoop.io.Writable;
-
-import edu.uci.ics.genomix.util.Marshal;
-
-public class ReadHeadSet extends TreeSet<ReadHeadInfo> implements Writable, Serializable {
+public class ReadHeadSet extends ExternalableTreeSet<ReadHeadInfo> implements Iterable<ReadHeadInfo> {
```
jakebiesinger: Seems like the `ExternalableTreeSet` should implement the `Iterable<T>` interface, yes?  It seems the `iterator` function doesn't actually make anything read-only, yes?  There is no such thing as `const` in Java... I think you can't really restrict in-place updates.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
```diff
@@ -1,88 +1,64 @@
...
 
     public int getOffsetFromReadId(long readId) {
-        for (ReadHeadInfo readHeadInfo : this) {
-            if (readHeadInfo.getReadId() == readId)
-                return readHeadInfo.getOffset();
+        ReadHeadInfo lowKey = ReadHeadInfo.getLowerBoundInfo(readId);
+        ReadHeadInfo highKey = ReadHeadInfo.getUpperBoundInfo(readId);
+        SortedSet<ReadHeadInfo> result = super.rangeSearch(lowKey, highKey);
+        if (result.size() != 0) {
+            for (ReadHeadInfo readHeadInfo : result) {
```
jakebiesinger: crazy.... so to query with an offset you have to iterate over the whole list?  I would say that an offset-based range query is the only way we will ever actually want to do a range query!

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ExternalableTreeSet.java
```diff
@@ -0,0 +1,247 @@
...
+     * Returns a view of the portion of this set whose elements range from fromElement, inclusive, to
+     * toElement, exclusive. (If fromElement and toElement are equal, the returned set is empty.)
+     * [lowKey, highKey)
+     * 
+     * @param lowKey
+     * @param highKey
+     * @return
+     */
+    public SortedSet<T> rangeSearch(T lowKey, T highKey) {
+        SortedSet<T> set = inMemorySet.subSet(lowKey, highKey);
```
jakebiesinger: looks like the SortedSet returned here is a view and not a copy.  Good deal :)

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
```diff
@@ -1,88 +1,64 @@
...
-        for (int i = 0; i < count; i++) {
-            ReadHeadInfo curInfo = new ReadHeadInfo(data, offset);
-            offset += curInfo.getLengthInBytes();
-            add(curInfo);
-        }
-        return offset;
+    /**
+     * @param data
+     * @param offset
+     * @return How many bytes read by this function.
```
jakebiesinger: mentioned elsewhere... in all the other implementations, this is the new offset, not the length

jakebiesinger: I'm okay to change it to length but we should be consistent.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
```diff
@@ -94,17 +70,77 @@ public String toString() {
...
         }
         sbuilder.append(']');
         return sbuilder.toString();
     }
 
-    public static ReadHeadSet getIntersection(ReadHeadSet list1, ReadHeadSet list2) {
-        ReadHeadSet intersection = new ReadHeadSet(list1);
-        intersection.retainAll(list2);
-        return intersection;
+    public void unionUpdate(ReadHeadSet readIds, float lengthFactor, boolean flipOffset, int otherLength) {
```
jakebiesinger: this function seems off-- you're not actually using `readIds`?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
```diff
@@ -94,17 +70,77 @@ public String toString() {
...
+            // stream theirs in, offset and flipped
+            for (ReadHeadInfo p : super.inMemorySet) {
+                int newPOffset = otherLength - 1 - p.getOffset();
+                this.add(p.getMateId(), p.getReadId(), (int) ((newPOffset + 1) * lengthFactor - lengthFactor),
+                        p.getThisReadSequence(), p.getMateReadSequence());
+            }
+        }
+    }
+
+    public void postAppendOffsets(int newThisOffset) {
```
jakebiesinger: this function seems to do the "prepend".  If I have two readids @ offsets 5 and 7 and I prepend some string to the total readset, say 10bp, the new offsets would be 15 and 17.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
```diff
@@ -94,17 +70,77 @@ public String toString() {
...
+            p.resetOffset(newThisOffset + p.getOffset());
+        }
+        super.isChanged = true;
     }
+
+    public void unionUpdate(ReadHeadSet setB) {
+        super.union(setB);
+    }
+
+    public void preAppendOffset(int newOtherOffset) {
```
jakebiesinger: and this one seems more like a flip than a "preAppend".  offsets 5 and 7 with a 10bp newOtherOffset become 5 and 3.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
```diff
@@ -94,17 +70,77 @@ public String toString() {
...
+    public Set<Long> getReadIdSet() {
+        HashSet<Long> set = new HashSet<Long>();
+        for (ReadHeadInfo p : super.inMemorySet) {
+            set.add(p.getReadId());
+        }
+        return set;
+    }
+
+    @Override
+    public ReadHeadInfo readEachElementFromDataStream(DataInput in) throws IOException {
```
jakebiesinger: seems this should be static and should be named `readFromDataStream` not `readEachElement`-- there's only one element coming back.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
```diff
@@ -94,17 +70,77 @@ public String toString() {
...
+        ReadHeadInfo info = new ReadHeadInfo();
+        info.readFields(in);
+        return info;
+    }
+
+    @Override
+    public void writeEachElementToDataStream(DataOutput out, ReadHeadInfo t) throws IOException {
```
jakebiesinger: same here-- static and only working with one element.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/util/TestUtils.java
```diff
@@ -161,16 +161,16 @@ private static boolean compareStringByFields(String lineExpected, String actualL
 
         for (int i = 0; i < unorderedFieldsAct.size(); i++) {
             ArrayList<String> subfieldsAct = new ArrayList<String>(Arrays.asList(unorderedFieldsAct.get(i)
-                    .split("\\s+")));
+                    .split("[^ATCG]+")));
```
jakebiesinger: this isn't used by any other function?  Can you rename the function to reflect the fact it only works on DNA strings?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/testhelper/BridgeAddVertex.java
```diff
@@ -41,7 +42,12 @@ public void initVertex() {
         if (kmerSize == -1) {
             kmerSize = 3; // Integer.parseInt(getContext().getConfiguration().get(GenomixJobConf.KMER_LENGTH));
         }
-        GenomixJobConf.setGlobalStaticConstants(getContext().getConfiguration());
+        try {
```
jakebiesinger: shouldn't you let these propagate?

jakebiesinger:
So there is no caching and range queries with offsets are very inefficient.  You're just moving the ReadHeadSet out of the vertex and into HDFS if it gets "too big".

I think that's fine for the sake of getting used to the provided API but hope to see some efficiency improvements :)

For the HDFS write, I wonder if pregelix exposes any API to tell if two nodes are on the same machine-- if they were, we could set a flag that would indicate "write to local disk" rather than HDFS.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/BTreeSet.java
JavierJia: Yep, if we use this BTree, we should pass it from the configuration. 
As we don't use it currently, we can keep it here .

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/BTreeSet.java
JavierJia: Done

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/Node.java
JavierJia: I implement here using the old master, before the newer changes. Let me change it to consistent with previous changes.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
JavierJia: Yes, I was hiding this API in the previous design. And the yesterday's scaffolding used the iterator() which not that obvious to me to make it as a separate API. I will hide this interface while understand more about that scaffolding code.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
JavierJia: I think the rangeSearch would be much faster than the previous iterating method.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ExternalableTreeSet.java
JavierJia: It said so. I think I can check their implemetation to see if we can also make a `view` of iterator

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
JavierJia: Yep, I'm merging the master yesterday. Let me change it.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
JavierJia: Thanks Jake, that is a bug. the `p` should loop throw the `readIds`.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
JavierJia: Good name!

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
JavierJia: Done.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
JavierJia: This weird code is to make the Generic Externalable<T> works. 
It's kind of hard for the Externalable<T> to get the class T . It needs the reflection to search for the class. And we need to set some extra things outside, which bring this low level data structure to the driver to get the hadoop or hyracks classloader.

And then I decide to move the creating newinstance job inside the `readFields` to the subclass, which don't need to load the context to new an instance.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
JavierJia: The same idea with the above `read`

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
jakebiesinger: I think scaffolding will need to do range queries but not iterate over the whole set.  If it's iterating currently, it's probably filtering as well (to do a range query!)

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
jakebiesinger: but as-is, you're still doing a linear scan.  The comparator should be comparing the offsets anyway, which would then make this code trivial and wouldn't need any scan at all.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/testhelper/BridgeAddVertex.java
JavierJia: I've propagate this to the genomix-pregelix, but it seems a lot of place called `initialIndex`, and that caused a lot of changes. 

Let me merge with the recent Anbang's #83 to check if I can change it easier.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
jakebiesinger: yeah creating `T` via new isn't really possible without using a `Factory` or reflection.  But still, seems like it should be static and should be called `readElement`.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
jakebiesinger: and this static as well and called `writeElement`.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
JavierJia: Let me check the usage of this function. 
This function actually doesn't make sense to me. A readId will correspond to a set of offset, this function should either `int getOffsetFromReadId(long readId, int offset)` or `Set<Integer> getOffsetFromReadId(long readId)`. ?

jakebiesinger: I think it's supposed to be "get me the offset of a readid with this id".  Meaning, I don't know what this id's offset is and I'd like to find out for some reason.  So I'm sure it's a single int being returned.

JavierJia: It was used in `BasicBFSTraverseVertex` only. 

```
    public int updateBFSLength(BFSTraverseMessage incomingMsg, UPDATELENGTH_TYPE type) {

        int totalBFSLength = incomingMsg.getTotalBFSLength();
        VertexValueWritable vertex = getVertexValue();
        int internalKmerLength = vertex.getInternalKmer().getKmerLetterLength();
        ReadHeadSet readHeadSet;
        int offset;
        switch (type) {
        // remember to account for partial overlaps
        // src or dest, as long as flip, updateLength += offset, otherwise
        // updateLength += kmerLength - offset
            case SRC_OFFSET:
                boolean srcIsFlip = incomingMsg.getSrcReadHeadOrientation() == READHEAD_ORIENTATION.FLIPPED;
                readHeadSet = srcIsFlip ? vertex.getFlippedReadIds() : vertex.getUnflippedReadIds();
                offset = readHeadSet.getOffsetFromReadId(incomingMsg.getReadId());
                return srcIsFlip ? offset : internalKmerLength - offset;
```

If you are sure there is only one offset for each readid, I'd rather use `ceiling` method of TreeSet here.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
JavierJia: The problem is that the  there is no `static abstract` method, isn't it ?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/util/TestUtils.java
JavierJia: done

JavierJia:
@jakebiesinger  please have a look again~

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
jakebiesinger: ah, another Java weirdness. fair enough.

jakebiesinger: still think the names should change for these.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/testhelper/BridgeAddVertex.java
jakebiesinger: if not, you can always wrap it in some other error, e.g., RuntimeException() or something. Better to error somehow than to fail silently.

jakebiesinger:
very minor comments left.  LGTM once they're addressed.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/testhelper/BridgeAddVertex.java
JavierJia: Done.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
JavierJia: I changed it to `readEachNonGenericElement`

JavierJia:
Any other comments ?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/type/ReadHeadSet.java
jakebiesinger: It doesn't matter all that much but the part of the name I took issue with is `each`.  Each makes it sound like you're reading a collection, which you're really not doing here!   I think I prefer `readElementFromDataStream` to `readEachNonGenericElement`.

But whatever, this can stay if you feel strongly about the current name.

JavierJia: :ok_hand: Let me change it in the next PR.

PR #92, opened 2013-12-07
Jianfeng/fix iterator to rangesearch

JavierJia:
@jakebiesinger @Elmira88 @anbangx 
This PR is to hide the iterator API in ReadHeadSet. I changed the corresponding voting branch code in RayVertex.


genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding/RayVertex.java
```diff
@@ -83,40 +87,23 @@ public void sendMsgToNeighbor(VKmer neighbor, EDGETYPE et) {
...
-        }
-    }
-
-    public int indexOfKmerOnRead(ReadHeadInfo readHead, VKmer kmer) {
-        int index;
-        String read = readHead.getThisReadSequence().toString();
-        String stringKmer = kmer.toString();
-        index = read.toLowerCase().indexOf(stringKmer.toLowerCase());
-        return index;
+        return readHead.getThisReadSequence().getGeneCodeAtPosition(neighborPositionOnRead) == neighbor
```
jakebiesinger: this should actually check if all K letters at this position are the same.  When I was describing this to you, I meant that the only information needed from the neighbor (in terms of sending a msg) was a single letter.  But the ReadHeadInfo should compare the full kmer, not just that single letter.

JavierJia: Yes, you are right. Please check my new fix~

jakebiesinger:
LGTM.  For paired-end reads (which will do a search across a large range rather than a single position), we should use @Nan-Zhang's recent substring search.

PR #94, opened 2013-12-08
Wbiesing/fix tests

jakebiesinger:
WIP to get all the travis test cases passing
- [ ] One test case fails but doesn't result in a nonzero exit code (see end of https://s3.amazonaws.com/archive.travis-ci.org/jobs/15107105/log.txt).
- [x] Need to fix fastq input format (zip it together into a readid file when uploading to HDFS?)
- [ ] Need to regenerate the expected outputs and **check to make sure they're correct**.  @anbangx do you want to handle this?


anbangx:
I will take care of regenerating the expected outputs and check to make sure they're correct.

jakebiesinger:
> I will take care of regenerating the expected outputs and check to make sure they're correct.

Thanks!  Go ahead and push to this branch when they're ready.

PR #96, opened 2013-12-08
add support for library-ids

jakebiesinger:
@Elmira88 @JavierJia @Nan-Zhang @anbangx 
This branch brings in support for separate libraries, allowing us to use more than one input file and to be able to specify insert size and read length information for each set of inputs.  For example, I may want to use as input 1) one single-end input file, 2) 2 paired-end input files (`_1` and `_2`) with insert size=200bp, and 3) paired-end inputs with insert size=1kb.

Several updates to deal with:
- to use the fastq input format, you must now use `-singleEndFastqs <library1> <library2> ...` and `-pairedEndFastqs <library1_mate1> <library1_mate2> <library2_mate1> <library2_mate2> ...`.
- if you specify `-pairedEndFastqs` or you are continuing a previous job and want the paired-end information in scaffolding, you must also specify `-outerDistanceMeans <library1_mean> <library2_mean> ...` and `-outerDistancesStdDev <library1_std> <library2_std> ...`.
- `ReadHeadInfo` now has been reorganized to include the `libraryId` of the read.  The offset also now takes up the most-significant bits so @JavierJia your comparisons in `ReadHeadSet` should be more natural.
- fastq files are now converted to readid format when uploading.  That means our parsers will never have to deal with them.

I need to do a bit more testing on this, but it will have to wait until Monday.


genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ReadHeadInfo.java
```diff
@@ -86,16 +93,26 @@ public void setUUID(long uuid) {
         value = uuid;
     }
 
-    public static long makeUUID(byte mateId, long readId, int posId) {
-        return (readId << 17) + ((posId & 0xFFFF) << 1) + (mateId & 0b1);
+    public static long makeUUID(byte mateId, byte libraryId, long readId, int offset) {
+        // check to make sure we aren't losing any information
+        if (mateId != (mateId & ~(-1 << bitsForMate)))
```
JavierJia, line 98: what's the advantage of using -1 here ?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ReadHeadInfo.java
```diff
@@ -127,19 +135,23 @@ public VKmer getMateReadSequence() {
     }
 
     public byte getMateId() {
-        return (byte) (value & 0b1);
+        return (byte) ((value & ~(-1 << (mateIdShift + bitsForMate))) >>> mateIdShift); // clear leading bits, then shift back to place
```
JavierJia, line 138: I think the -1 << 3 will keep the all higher bit as 1, and the last 3 bit as 000, then everything before `mateIdShift + bitsFormate` will keep as the original one. And then it will also return the previous non-mate bit also.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ReadHeadInfo.java
```diff
@@ -86,16 +93,26 @@ public void setUUID(long uuid) {
...
-        return (readId << 17) + ((posId & 0xFFFF) << 1) + (mateId & 0b1);
+    public static long makeUUID(byte mateId, byte libraryId, long readId, int offset) {
+        // check to make sure we aren't losing any information
+        if (mateId != (mateId & ~(-1 << bitsForMate)))
+            throw new IllegalArgumentException("byte specified for mateId will lose some of its bits when saved! (was: " + mateId + " but only allowed " + bitsForMate + " bits!");
+        if (libraryId != (libraryId & ~(-1 << bitsForLibrary)))
+            throw new IllegalArgumentException("byte specified for libraryId will lose some of its bits when saved! (was: " + libraryId + " but only allowed " + bitsForLibrary + " bits!");
+        if (readId != (readId & ~(-1 << bitsForReadId)))
+            throw new IllegalArgumentException("byte specified for readId will lose some of its bits when saved! (was: " + readId + " but only allowed " + bitsForReadId + " bits!");
+        if (offset != (offset & ~(-1 << bitsForLibrary)))
```
JavierJia: why `bitForLibrary` for offset ?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ReadHeadInfo.java
JavierJia, line 98: I don't think these checking lines make sense.
It should be 

```
if (mateId != (mateId << mateIdShift) >>> mateIdShift))
```

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ReadHeadInfo.java
```diff
@@ -86,16 +93,26 @@ public void setUUID(long uuid) {
...
+            throw new IllegalArgumentException("byte specified for offset will lose some of its bits when saved! (was: " + offset + " but only allowed " + offset + " bits!");
+        
+        return ((mateId << mateIdShift) + (libraryId << libraryIdShift) + (readId << readIdShift) + (offset << offsetShift));
```
JavierJia: if the order is
`mateId  libraryId  readid offset`
and the comparator just return Long, then we can't do the range query by `offset`.
I think it should be
`mateId libraryId offset readid`

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
```diff
@@ -356,6 +360,121 @@ public void runGenomix(GenomixJobConf conf) throws NumberFormatException, Hyrack
...
+     * If the filename ends with ".gz", the file is assumed to be in gzip format.
+     * The caller is responsible for closing the returned BufferedReader.
+     * 
+     * @throws IOException
+     */
+    private static BufferedReader openFile(String filename) throws IOException {
+        FileInputStream fis = new FileInputStream(filename);
+        BufferedReader br;
+        if (filename.endsWith(".gz")) {
+            InputStream gzipStream = new GZIPInputStream(fis); // do I need to explicitly close this reader?  Or will it be closed when the containing BufferedReader is closed?
```
JavierJia, line 466: It will close once the br is close().

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ReadHeadInfo.java
jakebiesinger, line 98: > if (mateId != (mateId << mateIdShift) >>> mateIdShift))

What I'm testing for here is that there are no significant bits lost.

Let's say I have a field with 3 bits: `0000 0XXX`.  The expression `~(-1 << numBitsInDataStructure)` gives a mask like `0000 0111` which when `&`'ed with the original value will clear all the significant bits.  If the cleared value `!=` the original value, then the passed argument is invalid and some bits would be lost (this is like an overflow flag).

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ReadHeadInfo.java
jakebiesinger, line 138: You're missing the `~` which flips all the bits in the field, leaving the higher bits as `0` and the last 3 bits as `1`.  You're right that the mask won't remove the bits that are less significant than the field we're trying to access, but the `>>>` at the end will drop anything less significant than the value.

For example, I have a field with shift=2 and numBits=3: `000X XX00`.  the mask for this will be `0001 1111` which will zero out the 3 most-significant bits, making `000X XX??`.  Then finally, the value is shifted to the right by 2, making `0000 0XXX`, which is the correct value to return.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ReadHeadInfo.java
jakebiesinger: Thanks, that's a bug.  fixed in next push.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ReadHeadInfo.java
jakebiesinger: The order of the internally stored value with most-significant bits first is: `offset library mate readid`.  The order these are added together doesn't matter but I'll go and change it to be clearer.

jakebiesinger: since the most-significant bits are the offset, you should be able to do the range query without any extra comparator (just compare the bits of the long value).

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
jakebiesinger, line 466: too bad there's still a java warning about the underlying GZip stream not being closed :/.  Think I should add a `@SuppressWarnings` to this statement?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ReadHeadInfo.java
JavierJia, line 138: Got it, :+1:

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ReadHeadInfo.java
JavierJia, line 98: Got it, I did not notice the '~'

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
JavierJia, line 466: I think so.

JavierJia:
LGTM :+1:

jakebiesinger:
Any other comments here?  We need this PR in place to do scaffolding properly.

JavierJia:
LGTM.

On Tue, Dec 10, 2013 at 4:16 PM, Jake Biesinger notifications@github.comwrote:

> Any other comments here? We need this PR in place to do scaffolding
> properly.
> 
> ?
> Reply to this email directly or view it on GitHubhttps://github.com/uci-cbcl/genomix/pull/96#issuecomment-30282505
> .

## 

---

Best Regards
Jianfeng

PR #97, opened 2013-12-09
recovered tests in genomix-data partially 

Nan-Zhang:
@jakebiesinger @JavierJia @anbangx @Elmira88 please have a check.
1. refactoring KmerFactoryTest is done while other types' tests is still on progress. If this KmerFactoryTest refactoring is reasonable, I can continue work on other tests.
2. removed EdgeMapTest, and cleaned duplicates tests in original KmerFactoryTest. 
3. Separate original KmerFactoryTest into two isolated test files called KmerFactoryFixedTest (corresponding to our original KmerFactoryTest which all use fixed test data as input), and KmerFactoryRandomTest(totally used random input data, increase the test random property and test pressure).
4. KmerFactoryRandomTest only need to set strMaxLength and strMinLength these 2 parameters in the setUp() function.After setting them, each test case will begin with a String "input" within the range(strMaxLength, strMinLength). Our testing input vkmer can be got by kmerFactory.getKmerByRead(kmerSize, input.getBytes(), 0) where kmerSize is determined by a range (1, input.length()) in a Random way.
5. Beside that, I added a class called RandomTestHelper which provide utility function for KmerFactoryRandomTest


JavierJia:
All those tests passed ? Then LGTM

jakebiesinger:
Yeah, thanks @Nan-Zhang.  Do we have tests in place for the readHeadInfo objects?  Particularly for when multiple `Node`s are merged together?  This LGTM but maybe that can be the next place for you to focus?

Nan-Zhang:
yeah, @jakebiesinger . According to this style, I will complete the separation of original Vkmer, VkmerList, Node, ReadHeadInfo, and ReadHeadSet test to fixed and random test class by this week. Especially for readHeadInfo and Node, I will add more tests when multiple Node are merged.

Nan-Zhang:
Yeah, @JavierJia  they are passed.

jakebiesinger:
> I will add more tests when multiple Node are merged.

Can you make this a priority?

Nan-Zhang:
yes. @jakebiesinger

jakebiesinger:
@Nan-Zhang are you going to merge this in?

Nan-Zhang:
@jakebiesinger no, I am updating this branch until I recover all genomix-data tests

jakebiesinger:
Okay, but it's hard for us to keep track of what's changed each time.  Jianfeng and I already gave the LGTM for this branch;  for the next round, we won't be able to look at the diff for this pull request and will instead have to do a manual comparison.

I think it's better to merge and close this pull request, then open another one for additional changes (even if it's the same branch name).

Nan-Zhang:
got it. @jakebiesinger  I think you are right. I will merge it first, and then make a new pull request.

PR #100, opened 2013-12-10
implement the simple lazy load

JavierJia:
@jakebiesinger 
Here is the lazy load implementation. The external set will not actually load the inMemorySet until some of the access method get called.


jakebiesinger:
Pretty straightforward. LGTM.

PR #101, opened 2013-12-12
refactor ray to work with a partially merged graph

jakebiesinger:
@Elmira88 @anbangx @JavierJia @Nan-Zhang 

Sorry this took a while to get the offsets correct for the paired-end reads.  In this branch, I've taken @Elmira88's code and refactored it pretty heavily to let the ray scaffolding work on a merged graph.  This is actually quite a bit more complicated than the unmerged case-- check out the score function for an example of how!

There are tons of `TODO`'s in the code.  I'll be chipping away at them, though anyone is welcome to contribute here.  There are several algorithmic questions to be addressed which I've raised in `TODO` comments as well.

@Nan-Zhang it would be great to have a version of your KMP algorithm that would let me specify ranges.  In the case of paired-end reads, we are searching for a particular subset of one kmer in a particular subset of another kmer.

I'd also love feedback on how to make the scoring more efficient, perhaps via some batch operations.


jakebiesinger:
At this point, the scaffolding works perfectly in an unmerged graph when the search is in the FORWARD direction :).  Hooray!

Merged graphs also seem to work but aren't as well tested (yet).  REVERSE search has a bug where the readids requested aren't present (we're looking in the wrong list).  I also haven't yet tested paired-end reads but the code is in place for them already.  I think there's likely a bug about which list is selected and possibly in getting the range queries right.  It took a lot of thinking to get them as they are and I'll probably have to work through the logic again with someone else to make sure about it.

I'm waiting on @anbangx's recent pull request about finding length and coverage cutoffs to implement `RayVertex.isStartSeed` properly.  It's also TBD how we handle initial search directions... Elmira previously had the search proceed in one direction until the seeds stopped growing.  I just read back over the Ray paper and they actually alternate, going one step in the FORWARD direction, then one step in the REVERSE direction, then F, R, F, R...

Still several TODO's in the code.

JavierJia:
Cool ! :+1: :cake: 
I'm still grading the TA course. Will be back on Monday.

Nan-Zhang:
Cool! @jakebiesinger do we need change the KMP?

jakebiesinger:
Yes, it would be great if we could change the kmp algorithm to work on
substrings of both the target and the query kmers.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/config/GenomixJobConf.java
```diff
@@ -612,14 +617,23 @@ public static void setGlobalStaticConstants(Configuration conf) throws IOExcepti
...
+            for (String length : conf.get(READ_LENGTHS).split(",")) {
+                readLengths.add(Integer.valueOf(length));
+            }
+        }
+        
         Byte libraryId = 0;
         if (conf.get(OUTER_DISTANCE_MEANS) != null) {
+            outerDistanceMeans = new HashMap<>();
             for (String mean : conf.get(OUTER_DISTANCE_MEANS).split(",")) {
                 outerDistanceMeans.put(libraryId++, Integer.valueOf(mean));
```
JavierJia, line 639: If the libraryId is just ++, why we use the `HashMap`?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayMessage.java
```diff
@@ -0,0 +1,303 @@
...
+        AGGREGATE_SCORE,
+        PRUNE_EDGE,
+        CONTINUE_WALK,
+        STOP;
+        public static final RayMessageType[] values = values();
+    }
+
+    /** for REQUEST_KMER, REQUEST_SCORE, AGGREGATE_SCORE, CONTINUE_WALK */
+    private VKmerList walkIds = null;
+    private ArrayList<Integer> walkOffsets = null;
```
JavierJia: Can you add some small explanation about this `walkOffsets`? 
Is it the offset of the Ids inside the `walkIds`?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayScores.java
```diff
@@ -0,0 +1,143 @@
...
+
+import edu.uci.ics.genomix.data.types.EDGETYPE;
+import edu.uci.ics.genomix.data.types.VKmer;
+import edu.uci.ics.genomix.data.types.VKmerList;
+
+public class RayScores implements Writable {
+
+    private HashMap<SimpleEntry<EDGETYPE, VKmer>, Rules> scores = new HashMap<>();
+    
+    public void addRuleCounts(EDGETYPE et, VKmer kmer, int ruleA, int ruleB, int ruleC) {
```
JavierJia, line 51: What's these `ruleA`, `ruleB`, `ruleC` ?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/config/GenomixJobConf.java
jakebiesinger, line 639: Some libraries aren't paired end so their library ids will be missing from
the list, leaving holes in the array.

Still, I've arranged this recently so that the paired end libraries always
come first. Plus, I could leave the missing entries as null values.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayMessage.java
jakebiesinger: That's exactly right. The nodes identified by the walkids are offset in the
graph by walkoffsets.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayScores.java
jakebiesinger, line 51: These are three heuristics adapted directly from the Ray paper. There's a
short explanation about them in the RayScores class but I can add more.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayMessage.java
```diff
@@ -0,0 +1,303 @@
...
+    private Integer walkLength = null;
+    
+    /** for REQUEST_KMER, PRUNE_EDGE, CONTINUE_WALK */
+    private EDGETYPE edgeTypeBackToFrontier = null;
+    
+    /** for REQUEST_KMER, CONTINUE_WALK */
+    private Boolean frontierFlipped = false;
```
anbangx: what's frontierFlipped and edgeTypeBackToFrontier? Add some explanation?

jakebiesinger: I'll add some explanation in the source.  Basically, there are three node types: 1) the "frontier", where the decision about which branch to take/prune is being made, 2) the "candidates" which are attached to to frontier and whose kmers are being evaluated, and 3) the "walk" nodes, which are previously visited frontiers and contribute their readids to the decision in the current frontier.

`edgeTypeBackToFrontier` is the edge type from the particular candidate represented by this message back to the frontier.
`frontierFlipped` stores whether or not the frontier node is in the same orientation as the `INITIAL_SEARCH_DIRECTION`.

jakebiesinger:
> Cool ! :cake: 

The cake is a lie..?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayMessage.java
jakebiesinger: Done.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayMessage.java
jakebiesinger: Done.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayScores.java
jakebiesinger, line 51: Done.

jakebiesinger:
Hooray!  The algorithm now works in unmerged graphs in both directions!

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
```diff
@@ -0,0 +1,544 @@
...
+                        - GenomixJobConf.outerDistanceStdDevs.get(libraryId));
+            }
+        }
+        MAX_DISTANCE = Math.max(MAX_OUTER_DISTANCE, MAX_READ_LENGTH);
+    }
+
+    @Override
+    public void compute(Iterator<RayMessage> msgIterator) throws Exception {
+        if (getSuperstep() == 1) {
+            initVertex();
```
anbangx: Why do you need this initVertex() here? You already add it in confgure.

jakebiesinger: True.  good catch.  I'll remove it.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
```diff
@@ -0,0 +1,544 @@
...
+                case CONTINUE_WALK:
+                    startBranchComparison(msg);
+                    break;
+                case REQUEST_KMER:
+                    sendKmerToWalkNodes(msg);
+                    break;
+                case REQUEST_SCORE:
+                    // batch-process these (have to truncate to min length)
+                    RayMessage reqMsg = new RayMessage();
+                    reqMsg.setAsCopy(msg);
```
anbangx: This can be simplified to requestScoreMsgs.add(new RayMessage(msg));

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
```diff
@@ -0,0 +1,544 @@
...
+                    // so that other walks are free to include me
+                    vertex.getEdges(msg.getEdgeTypeBackToFrontier()).remove(msg.getSourceVertexId());
+                    vertex.visited = false;
+                    break;
+                case STOP:
+                    // I am a frontier node but one of my neighbors was already included in a different walk.
+                    // that neighbor is marked "intersection" and I need to remember the stopped state.  No path
+                    // can continue out of me in the future, even when I receive aggregate scores
+                    vertex.stopSearch = true;
+                    // TODO obey this stop search in the CONTINUE branch
```
anbangx, line 174: Some questions: 
1. If there are branches and they are in the range, does the prune always keep one of them and remove others? Or sometimes it doesn't prune any branch?
2. I am thinking if you mark all the walk nodes as visited and stop without any successful pruning. All this walks nodes are visited and can't be continued out by other paths?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
```diff
@@ -0,0 +1,544 @@
...
+                    break;
+            }
+        }
+        if (requestScoreMsgs.size() > 0) {
```
anbangx, line 178: Why do you prefer to put this code here rather than inside case block?

jakebiesinger, line 178: these two functions require the entire set of messages, not just one.  The other message types can be dealt with using an iterator; these two need the whole list.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
jakebiesinger: True.  I'll change that.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
```diff
@@ -0,0 +1,544 @@
...
+        // must explicitly check if this node was visited already (would have happened in 
+        // this iteration as a previously processed msg!)
+        if (vertex.visited) {
+            vertex.intersection = true;
+            vertex.stopSearch = true;
+            return;
+        }
+        vertex.visited = true;
+        // if the prev frontier was flipped and I came across a FR/RF, I'm back to unflipped
+        vertex.flippedFromInitialDirection = msg.isCandidateFlipped();
```
anbangx, line 201: isCandidateFlipped()? From my understanding, it's not only candidate will receive this CONTINUE_WALK message. 
Is it isWalkNodeFlipped()?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
```diff
@@ -0,0 +1,544 @@
...
+        DIR nextDir = msg.getEdgeTypeBackToFrontier().mirror().neighborDir();
+
+        msg.visitNode(id, vertex);
+
+        if (vertex.degree(nextDir) == 0) {
+            // this walk has reached a dead end!  nothing to do in this case.
+            return;
+        } else if (vertex.degree(nextDir) == 1) {
```
anbangx, line 210: If this is for merged graph, how can this happen?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
jakebiesinger, line 174: > 1. If there are branches and they are in the range, does the prune always keep one of them and remove others? Or sometimes it doesn't prune any branch?

Sometimes, it won't prune any path.  Specifically, any time there isn't a single branch that dominates _all_ the other branches, the walk will stop completely.

> 1. I am thinking if you mark all the walk nodes as visited and stop without any successful pruning. All this walks nodes are visited and can't be continued out by other paths?

This would only apply to the very first node (where the only walk node is the frontier node) or to walks that only saw path nodes and then failed to prune. We could special-case this and send a message back to the walk nodes to "unvisit" them.  Not sure it's worth it though, especially in a merged graph where there aren't very many path nodes in the graph.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
```diff
@@ -0,0 +1,544 @@
...
+            outgoingMsg.reset();
+            outgoingMsg.setMessageType(RayMessageType.STOP);
+            sendMsg(msg.getSourceVertexId(), outgoingMsg);
+            return;
+        }
+
+        outgoingMsg.reset();
+        outgoingMsg.setMessageType(RayMessageType.REQUEST_SCORE);
+        outgoingMsg.setSourceVertexId(msg.getSourceVertexId()); // frontier node
+        outgoingMsg.setFrontierFlipped(msg.getFrontierFlipped());
```
anbangx, line 255: in sendScoresToFrontier, I don't see you use FrontierFlipped and EdgeTypeBackToFrontier. Why do you send them through message?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
```diff
@@ -0,0 +1,544 @@
...
+                outgoingMsg.setWalkLength(walkLength);
+                outgoingMsg.getWalkOffsets().add(myOffset);
+                outgoingMsg.getWalkIds().append(id);
+                sendMsg(frontierNode, outgoingMsg);
+            }
+            return;
+        }
+
+        // get the smallest kmer in all the messages I've received
+        int minLength = msgs.get(0).getWalkIds().getPosition(0).getKmerLetterLength();
```
anbangx: Add comments to explain why need this minLength.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
jakebiesinger, line 201: In this case, I am a new frontier node and this message is coming from the previous frontier.  So from the message's perspective, I am the candidate referred to by this function.  I'll add a comment clarifying this.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
```diff
@@ -0,0 +1,544 @@
...
+        for (int i = 1; i < msgs.size(); i++) {
+            minLength = Math.min(minLength, msgs.get(i).getWalkIds().getPosition(0).getKmerLetterLength());
+        }
+        minLength = minLength - Kmer.getKmerLength() + 1;
+
+        // I'm now allowed to score the first minLength kmers according to my readids
+        RayScores singleEndScores = voteFromReads(true, vertex, msgs, myOffset, walkLength, minLength);
+        RayScores pairedEndScores = HAS_PAIRED_END_READS ? voteFromReads(false, vertex, msgs, myOffset, walkLength,
```
anbangx, line 315: For pairedEnd, both singleEndScores and pairedEndScores are used? Or only pairedEndScores?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
jakebiesinger, line 210: path nodes can still exist in the merged graph.  Think of simple bubble nodes-- they are path nodes!

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
jakebiesinger, line 255: They're not needed in this step but they are needed in the next step.  I think it's clearer to put them in the message and pull them out than to put them in the vertex value and wait around for the messages to show up.  Maybe it would be clearer the other way though?  I'm NOT worried about the 16 bits extra per message here though :)

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
jakebiesinger, line 315: If a library is paired-end, it will calculate BOTH single and pairedend scores.  The pairedend scores are checked for dominance first.  If they don't dominate, the singleend scores are checked.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
jakebiesinger: Will do.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
```diff
@@ -0,0 +1,544 @@
...
+     * use the reads present in this vertex to score the kmers in this message list.
+     * For each message,
+     * | for each read in a subset of reads in this vertex
+     * | | run a sliding window of length (original) Kmer.length
+     * | | | see if all the letters in the sliding window match the read
+     * So somehow, we've turned this into a n**4 operation :(
+     * // TODO for single-end reads, we could alternatively count how many letters in the VKmers match
+     * // or we could base the score on something like edit distance
+     */
+    private static RayScores voteFromReads(boolean singleEnd, RayValue vertex, ArrayList<RayMessage> msgs,
```
anbangx: Add example for this... Have no idea what it does.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
```diff
@@ -0,0 +1,544 @@
...
+            walkIds.append(msg.getWalkIds().getPosition(0));
+            walkOffsets.add(msg.getWalkOffsets().get(0));
+        }
+
+        VKmer dominantKmer = null;
+        EDGETYPE dominantEdgeType = null;
+        // need to compare each edge in this dir with every other one.  Unfortunately, this is ugly to do, requiring 
+        // us to iterate over edge types, then edges in those types, and keeping track of the indexes ourselves, etc :(
+        // this 4 loops are really just two for loops that are tracking 1) the index, 2) the edge type, and 3) the kmer
+        int queryIndex = 0, targetIndex = 0;
```
anbangx, line 565: Also add examples for this.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
jakebiesinger: Done.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
jakebiesinger: Done.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
jakebiesinger, line 201: Should be clearer now.  Check it out.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
jakebiesinger: Done.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
jakebiesinger: I added a fairly extensive scoring example to the docstring of this method.  Check it out and give us some feedback :)

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
jakebiesinger, line 565: not sure what you wanted here but hopefully the example at the header of this method is sufficient to explain...

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
anbangx, line 174: Especially in a merged graph, if there is no successful prune and you mark it as visited and other paths can't continue out of it, is it correct?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
anbangx, line 565: I want the example how you compute the score

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
jakebiesinger, line 174: Yeah, that's correct.  But it's for a good reason.  Perhaps you don't prune anything in this iteration but you were able to prune many things in previous iterations.  Previous nodes in the walk shouldn't be pruned off by some other walk, right?  What happens when another walk comes in to nodes that you've already visited?  should it be able to prune the edges that you just walked across?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/scaffolding2/RayVertex.java
jakebiesinger, line 565: I added one.  Look at the `voteFromReads` method's docstring.  There's a non-trivial example there.

jakebiesinger:
@anbangx I improved the examples a bit and added an explanation about why we must do all-vs-all comparison.

jakebiesinger:
I think there's a regression in path merge.  In my testing, the reads don't have the correct offsets in a merged graph :(

jakebiesinger:
I just ran this on some synthetic data. The graph was merged beforehand. From one start seed that I chose manually (decent length and # of reads), this algorithm chose 11 dominant edges over the course of 60 iterations before reaching a dead end and stopping.  There were normally 2-4 path nodes between each prune.

I also ran pathmerge after ray scaffolding: it ran for only six iterations (up to 3 merges) and I'm not sure how it affected the overall length distribution.

There are still lots of optimizations to do here but I think it's almost time to merge this in.  Can you guys take another look and give feedback?

jakebiesinger:
Items left to do before merging in (Please edit this comment directly to add your own items):
- [x] set the INITIAL-DIRECTION appropriately
- [x] implement isStartSeed() based on @anbangx's #109 
- [x] obey STOP message in prune

jakebiesinger:
There are also several "to think about" items:
- what does "ruleC" mean when the graph has been merged already?
- can we have the search proceed in both directions simultaneously (have to keep walk orientation as part of the messages, have to group request messages by their frontiers)
- how can we effectively prune the "back edges"?  How can we do so without losing the chance to connect with other walks?  Perhaps the walk that's followed is actually more important than the mutations to the graph? Having multiple _walkers_ rather than _pruners_ might be interesting... and I wonder if that's how Ray is actually proceeding...

jakebiesinger:
And a couple of optimizations that could happen:
- separate ReadHeadSet's for each library.  This would make querying more efficient (right now we have to use the `max(readLengths)` to determine where to query but libraries have different lengths... some may become irrelevant faster than others; those that aren't relevant wouldn't have to be queried
- the scoring function is VERY inefficient (like `O(N**4)`).  It would be cool to have a more efficient way (#106 is helpful but I think we could do even better, especially where we are looking for exact matches rather than matches in a range.
- storage of the ReadHeadSet is pretty inefficient.  You have tons of redundant information in that data structure (most reads will be very similar to each other).  It would be cool to see a more efficient data structure there.
- might make sense to reorder the incoming messages to make the process a little more deterministic

jakebiesinger:
@anbangx can you implement the TODO "implement isStartSeed() based on @anbangx's #109" ?  I'm still not sure what that entails.

anbangx:
Sure.

On Tue, Dec 17, 2013 at 11:14 PM, Jake Biesinger
notifications@github.comwrote:

> @anbangx https://github.com/anbangx can you implement the TODO
> "implement isStartSeed() based on @anbangx https://github.com/anbangx's
> #109 https://github.com/uci-cbcl/genomix/pull/109" ? I'm still not sure
> what that entails.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/uci-cbcl/genomix/pull/101#issuecomment-30821455
> .

## 

Best Regards,

Anbang Xu

jakebiesinger:
Tick the box above when you're done :)

PR #105, opened 2013-12-13
Anbangx/fitting mixture model

anbangx:

This PR adds -setCutoffCoverageByFittingMixture and - setMinScaffoldingSeedLength two command line options to automatically generate minCoverage and minScaffoldingSeedLength parameters.

The minCoverage generation is based on fitting our data into mixture model(Exponential and Normal).
Assume there are two groups (an exponential and a gaussian), every point has some probability of belonging to each category and each category has an overall probabbility, first calculating the means of the exponential and gaussian, considering every point but weighting it by the pdf value for that point in that class. The weighted means and variances are the new parameters, which gives you new pdf values. Repeat to run the whole steps until the likelihood converges. Actually this is the EM algorithm. And then one choice of our coverage threshold can be when the probability of belonging to the normal distribution is greater than the probability of belonging to the exponential distribution, this happens when the ratio of norm to exponential is > 1
I have tested it. It works well.

The minScaffoldingSeedLength generation is based on the goal that we want to get the top 1%(top means top kmerLength) of our vertices as our scaffolding seed.


anbangx:
This PR is separated into #107 and #109 .

PR #106, opened 2013-12-16
change KMP to support query in range

Nan-Zhang:
@jakebiesinger  @JavierJia @anbangx @Elmira88  please have a look. This branch is for modifying KMP to support range query which searches a particular subset of src_kmer whether is contained in a particular subset of targt_kmer. if it contain this subset, return the index, else return -1. sinlge test already been passed.


jakebiesinger:
Hi @Nan-Zhang it seems you've been pushing to this branch several extra commits?  isn't 78f0830 the last commit that should be in this branch?

jakebiesinger:
If that's the case, please do the following:

```
git checkout nanzhang/KMP_Range
git reset --hard 78f0830
git push -f origin nanzhang/KMP_Range
```

which will reset this branch to commit 78f0830. You can then get those other changes in a new branch via:

```
git checkout genomix/fullstack_genomix
git pull
git checkout -b nanzhang/fixing-kmer-or-whatever-name
git cherry-pick  78f0830..1955e1a
```

which will create a new branch based on the main branch and will apply the other commits on top of it.

Nan-Zhang:
@jakebiesinger actually I pushed some recover tests to this wrong branch. Thanks jake. I learned it.

jakebiesinger:
> Thanks jake. I learned it.

Can you fix this branch so it doesn't have those recovery tests?  follow my instructions above, then do a force-push via `git push -f origin nanzhang/KMP_Range`

Nan-Zhang:
@jakebiesinger  already updated it:)

PR #107, opened 2013-12-17
Anbangx/find start seed min length

anbangx:

This PR is for findMinScaffoldingMinLength, which separates the setCutoffCoverage and findMinScaffoldingMinLength from the previous PR I sent. hpc is broken, I need more time to test that Gaussian model in the real data. So this PR can be used to get startSeed first.

### DIFF
--- a/genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/config/GenomixJobConf.java
+++ b/genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/config/GenomixJobConf.java
@@ -187,8 +187,8 @@ public class GenomixJobConf extends JobConf {
         @Option(name = "-runAllStats", usage = "Whether or not to run a STATS job after each normal job")
         private boolean runAllStats = false;
         
-        @Option(name = "-logBubbleInfo", usage = "Whether or not to log bubble info on simple bubble merge")
-        private boolean logBubbleInfo = false;
+        @Option(name = "-setMinScaffoldingSeedLength", usage = "Whether or not to automatically set min scaffolding seed length")
+        private boolean setMinScaffoldingSeedLength = false;
     }
 
     /**
@@ -287,6 +287,8 @@ public class GenomixJobConf extends JobConf {
     public static final String HDFS_WORK_PATH = "genomix.hdfs.work.path";
     public static final String EXTRA_CONF_FILES = "genomix.conf.extraConfFiles";
     public static final String RUN_ALL_STATS = "genomix.conf.runAllStats";
+    public static final String SET_MIN_SCAFFOLDING_SEED_LENGTH = "genomix.conf.setMinScaffoldingSeedLength";
+    public static final String MIN_SCAFFOLDING_SEED_LENGTH = "genomix.conf.minScaffoldingSeedLength";
 
     // Graph cleaning   
     public static final String BRIDGE_REMOVE_MAX_LENGTH = "genomix.bridgeRemove.maxLength";
@@ -582,6 +584,7 @@ public class GenomixJobConf extends JobConf {
         if (opts.extraConfFiles != null)
             set(EXTRA_CONF_FILES, opts.extraConfFiles);
         setBoolean(RUN_ALL_STATS, opts.runAllStats);
+        setBoolean(SET_MIN_SCAFFOLDING_SEED_LENGTH, opts.setMinScaffoldingSeedLength);
     }
 
     /**
--- a/genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/config/GenomixJobConf.java
+++ b/genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/config/GenomixJobConf.java
@@ -186,9 +186,6 @@ public class GenomixJobConf extends JobConf {
 
         @Option(name = "-runAllStats", usage = "Whether or not to run a STATS job after each normal job")
         private boolean runAllStats = false;
-        
-        @Option(name = "-setMinScaffoldingSeedLength", usage = "Whether or not to automatically set min scaffolding seed length")
-        private boolean setMinScaffoldingSeedLength = false;
     }
 
     /**
@@ -287,7 +284,6 @@ public class GenomixJobConf extends JobConf {
     public static final String HDFS_WORK_PATH = "genomix.hdfs.work.path";
     public static final String EXTRA_CONF_FILES = "genomix.conf.extraConfFiles";
     public static final String RUN_ALL_STATS = "genomix.conf.runAllStats";
-    public static final String SET_MIN_SCAFFOLDING_SEED_LENGTH = "genomix.conf.setMinScaffoldingSeedLength";
     public static final String MIN_SCAFFOLDING_SEED_LENGTH = "genomix.conf.minScaffoldingSeedLength";
 
     // Graph cleaning   
@@ -584,7 +580,6 @@ public class GenomixJobConf extends JobConf {
         if (opts.extraConfFiles != null)
             set(EXTRA_CONF_FILES, opts.extraConfFiles);
         setBoolean(RUN_ALL_STATS, opts.runAllStats);
-        setBoolean(SET_MIN_SCAFFOLDING_SEED_LENGTH, opts.setMinScaffoldingSeedLength);
     }
 
     /**
--- a/genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
+++ b/genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
@@ -140,6 +140,7 @@ public class GenomixDriver {
             sortedCounterArray[Integer.parseInt(c.getName()) - minLength] = c.getValue();
         }
         
+        sortedCounter.clear();
         for(int i = 0; i < sortedCounterArray.length; i++)
             sortedCounter.add(sortedCounterArray[i]);
         return maxLength;
--- a/genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
+++ b/genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
@@ -36,6 +36,7 @@ import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.mapred.Counters;
 import org.apache.hadoop.mapred.Counters.Counter;
+import org.apache.hadoop.mapred.Counters.Group;
 import org.apache.hadoop.mapred.FileInputFormat;
 import org.apache.hadoop.mapred.FileOutputFormat;
 import org.apache.hadoop.mapred.JobConf;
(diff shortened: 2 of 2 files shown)


genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
```diff
@@ -99,6 +98,28 @@ private void setOutput(GenomixJobConf conf, Patterns step) {
...
+        float percentage = 0.2f; // TODO Should it provide by user?
+        long numOfSeeds;
+        // minimum number of nodes is 100
+        if (Math.abs((float) totalNodes * percentage) < 100) 
+            numOfSeeds = (long) Math.abs((float) totalNodes * percentage);
+        else
+            numOfSeeds = 100;
+
+        long curNumOfSeeds = 0;
+        for (Counter c : counters.getGroup("kmerLength-bins")) {
```
jakebiesinger: Are you sure this will return the values in order?  Seems like this algorithm relies on the bins being ordered (so small ones come first, right?).

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
```diff
@@ -99,6 +98,28 @@ private void setOutput(GenomixJobConf conf, Patterns step) {
         FileOutputFormat.setOutputPath(conf, new Path(curOutput));
     }
 
+    private void setMinScaffoldingSeedLength(GenomixJobConf conf) throws IOException {
+        Counters counters = GraphStatistics.run(curOutput, curOutput + "-kmerLength-stats", conf);
+        long totalNodes = counters.getGroup("totals").getCounter("nodes");
+        float percentage = 0.2f; // TODO Should it provide by user?
```
jakebiesinger: Yeah, 20% is way too high.  I think the value that we'd like to adjust would be to be able to specify the start seeds... but maybe we can just overload (again) the `debugKmers`?

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
```diff
@@ -99,6 +98,28 @@ private void setOutput(GenomixJobConf conf, Patterns step) {
...
+        long numOfSeeds;
+        // minimum number of nodes is 100
```
jakebiesinger: Doesn't this code have a _maximum_ of 100?

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
jakebiesinger: this should also be "fraction" not "percentage"

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
```diff
@@ -121,6 +142,9 @@ private void addStep(GenomixJobConf conf, Patterns step) throws Exception {
             case MERGE:
             case MERGE_P4:
                 pregelixJobs.add(P4ForPathMergeVertex.getConfiguredJob(conf, P4ForPathMergeVertex.class));
+                flushPendingJobs(conf);
+                if (Boolean.parseBoolean(conf.get(GenomixJobConf.SET_MIN_SCAFFOLDING_SEED_LENGTH)))
```
jakebiesinger: You want it to run automatically after every merge job?  Interesting... I guess it will always be up-to-date.

Maybe it would make sense to integrate this step directly into the merge phase? Meaning we could keep length counters there?

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
anbangx: Yeah, should add a sorter here.

jakebiesinger:
One major problem here: we need a conditional the length distribution!  We want the top 1% conditional on their coverage being `>` some min threshold.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/config/GenomixJobConf.java
```diff
@@ -187,8 +187,8 @@
         @Option(name = "-runAllStats", usage = "Whether or not to run a STATS job after each normal job")
         private boolean runAllStats = false;
         
-        @Option(name = "-logBubbleInfo", usage = "Whether or not to log bubble info on simple bubble merge")
-        private boolean logBubbleInfo = false;
+        @Option(name = "-setMinScaffoldingSeedLength", usage = "Whether or not to automatically set min scaffolding seed length")
```
jakebiesinger: If you don't set it, the value will be `null`, right?

anbangx: setMinScaffoldingSeedLength is asking user if they want to set this minlength. Maybe it is always needed? It doesn't need this anymore?

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
anbangx: overload (again) the debugKmers?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/config/GenomixJobConf.java
jakebiesinger: but there's no corresponding option that allows the user to set it?

Maybe you're missing a commit?  I only see two here but you mentioned there were three incoming?

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
jakebiesinger: In other words, if the user specifies `-debugKmers ACTACGTGAC,ACGTACGTGTGC` then you could assume those were the seeds you're interested in.  Guess that's my job (in scaffolding), not yours :)

anbangx: I guess this should be additional startSeeds. We also need to provide this minLength to cut startSeed.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/config/GenomixJobConf.java
anbangx: you mean setBoolean(SET_MIN_SCAFFOLDING_SEED_LENGTH, opts.setMinScaffoldingSeedLength);?

in 587 line

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
jakebiesinger: I think you could either do "in addition to" or "in place of".

anbangx: Yeah, it's better. And for this fraction, should it be provided by user?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/config/GenomixJobConf.java
jakebiesinger: There's no way for the user to directly specify `MIN_SCAFFOLDING_SEED_LENGTH`.  They can only specify whether or not to have it set after P4, right?  There's no way for them to specify it on the cmd line?

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
jakebiesinger: You think "in addition to" is better than "in place of"?  In other words, if the user specifies `-debugKmer`, we would only use those kmers as start seeds?

I don't know that the fraction needs to be specified by the user.  Maybe it would be helpful for now?  but not in the long term...

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
anbangx: I don't follow. Why is there this 100 minimun number of nodes? If you want to filter 80 nodes which less than 100, you filter 100.

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
anbangx: No. I think if we can provide both "in addition to" and "in place of" is better.

So what's the best number for this fraction?

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
jakebiesinger: ?  isn't this equivalent to:

```
numSeeds =  totalNodes * percentage;
if (numSeeds >= 100)
    numSeeds = 100;
```

jakebiesinger: Also, what's with the `Math.abs`?

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
jakebiesinger: > No. I think if we can provide both "in addition to" and "in place of" is better.

You want to be able to do both?  Yet more cmdline switches??

> So what's the best number for this fraction?

I would place it somewhere around 1% or as you're using fractions here, `.001`.

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
anbangx: Yeah, integrating this step into merge phase can save one more step to do hadoop runStats.

But we need to make sure we get this minLength before starting scaffolding.

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
anbangx: here, numSeeds may get 10.4

jakebiesinger: > here, numSeeds may get 10.4

but `Math.abs` will return all-positive values.  Maybe you want `Math.round`? or `Math.floor`?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/config/GenomixJobConf.java
anbangx: Remove!

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
anbangx: For now, how about only doing "in addition to"?

I set it as 1%.

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
anbangx: Fixed!

anbangx:
@jakebiesinger Refactor based on your feedback.

PR #108, opened 2013-12-17
add to local options

JavierJia:
@jakebiesinger 
I just add one option to store the external file in the local file system or the Hdfs. 
By default, it will store on the hdfs. The option should be specified in the constructor.

### DIFF
--- a/genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ExternalableTreeSet.java
+++ b/genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ExternalableTreeSet.java
@@ -8,13 +8,11 @@ import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.OutputStream;
 import java.io.Serializable;
-import java.net.InetAddress;
-import java.text.SimpleDateFormat;
-import java.util.Date;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.SortedSet;
 import java.util.TreeSet;
+import java.util.UUID;
 
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
@@ -314,8 +312,6 @@ public abstract class ExternalableTreeSet<T extends WritableComparable<T> & Seri
             return conf;
         }
 
-        protected static final SimpleDateFormat simpleDateFormat = new SimpleDateFormat("ddMMyy-hhmmssSS");
-
         public Path createFile(boolean local) throws IOException {
             if (local) {
                 return createOneFile(allocatedLocalPath, lfs, localWorkPath);
@@ -328,10 +324,7 @@ public abstract class ExternalableTreeSet<T extends WritableComparable<T> & Seri
                 Path workPath) throws IOException {
             Path path;
             do {
-                String hostName = "";
-                hostName = InetAddress.getLocalHost().getHostName();
-                path = new Path(workPath, ExternalableTreeSet.class.getName() + hostName
-                        + simpleDateFormat.format(new Date()));
+                path = new Path(workPath, "ExternalableTreeSet" + UUID.randomUUID());
             } while (fs.exists(path));
             validation.put(path, fs.create(path, (short) 1));
             return path;
--- a/genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ExternalableTreeSet.java
+++ b/genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ExternalableTreeSet.java
@@ -8,6 +8,7 @@ import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.OutputStream;
 import java.io.Serializable;
+import java.nio.file.Files;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.HashMap;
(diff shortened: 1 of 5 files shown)


genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ExternalableTreeSet.java
```diff
@@ -51,9 +51,15 @@ public static void setCountLimit(int count) {
...
     public ExternalableTreeSet() {
+        this(false);
+    }
+
+    public ExternalableTreeSet(boolean isLocal) {
         this(null);
+        this.isLocal = isLocal;
     }
 
     protected ExternalableTreeSet(Path path) {
```
jakebiesinger: No constructor using (Path, boolean)?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ExternalableTreeSet.java
```diff
@@ -243,79 +251,117 @@ public void write(DataOutput out) throws IOException {
...
-            workPath = workingPath;
-            allocatedPath = new HashMap<Path, OutputStream>();
+        private HashMap<Path, OutputStream> allocatedHdfsPath;
+        private HashMap<Path, OutputStream> allocatedLocalPath;
+
+        public FileManager(Configuration conf, Path hdfsWorkingPath) throws IOException {
+            hfs = FileSystem.get(conf);
+            lfs = FileSystem.getLocal(conf);
+            hdfsWorkPath = hdfsWorkingPath;
+            localWorkPath = new Path(System.getProperty("java.io.tmpdir"));
```
jakebiesinger: It might make sense to dump this into the hyracks or pregelix tmp directory instead.

jakebiesinger:
those are minor comments; otherwise, LGTM.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ExternalableTreeSet.java
JavierJia: I change it to `ExternalableTreeSet(Path path, boolean local)`

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ExternalableTreeSet.java
JavierJia: Do you know how to get the hyracks tmp directory ?

jakebiesinger: Actually, in our case it's set up by the startCluster.sh script. The Hyracks nc directory is `conf.get("WORKPATH") + "/hyracks/nc"` and the pregelix one is `conf.get("WORKPATH") + "/pregelix/nc"`.

You might also be able to get the pregelix one via `pregelix.ClusterConfig` but I'm not sure...

jakebiesinger:
It would be nice if the locality of the data were tied to the object rather than the stream.  In particular, if I create an object using `isLocal=true` which is then read from a stream, I expect to have the data _written_ to the local filesystem, even if the `readFields` stream is HDFS.  Likewise, for #117, I need to be able to specify when the maximum count should be overridden completely and the data should be written entirely to the `write` output stream.

If I can't have that interface, then at least update the `setAsCopy`... you don't account for cases where the other object is in HDFS but you're supposed to write locally, nor vice-versa. The easiest solution might be to just force a `loadInMemorySetFromPath()` call in the `setAsCopy` code, which I can then directly call.  But I'd much prefer the method in the first paragraph.

JavierJia:
@jakebiesinger , please have a look. 
1. Split the readFrom and WriteTo setting. 
2. Add the forceWriteEntireBody() to write the whole body instead of the file.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ExternalableTreeSet.java
```diff
@@ -234,89 +248,136 @@ public void readFields(DataInput in) throws IOException {
...
+            }
+            allocatedLocalPath.clear();
         }
 
         public Configuration getConfiguration() {
             return conf;
         }
 
-        protected final SimpleDateFormat simpleDateFormat = new SimpleDateFormat("ddMMyy-hhmmssSS");
+        protected static final SimpleDateFormat simpleDateFormat = new SimpleDateFormat("ddMMyy-hhmmssSS");
```
jakebiesinger: regarding this output format, in HDFS, couldn't you have different NC's with the exact same file?  You use a lock which prevents threads within this NC from using the same time but different NCs could still get the same time.  

You could incorporate the hostname via a locked, static allocation block which calls to `DriverUtils.getIP("localhost");`.  In other words, incorporate the hostname into the filename.  Then the file will be guaranteed unique.

jakebiesinger: Actually, we should be able to do this by a simple system call rather than the ssh followed by a system call...

JavierJia: Yep, I just tried that bash, and it gave me some error.
I add that host part, and I think that won't cause too much problem, because I run a do while () there to make sure the create the different file.

jakebiesinger: Okey doke.

jakebiesinger:
LGTM

jakebiesinger:
actually just a sec

jakebiesinger:
LGTM

PR #109, opened 2013-12-18
Anbangx/test cutoff coverage

anbangx:

I just finished the test for the cutoff coverage on big dataset. It works fine now. Have a look at it. 

This PR is based on @jakebiesinger 's Ray-partially-merged-graph branch.


jakebiesinger:
No major changes since last time we discussed this it seems?

anbangx:
No major changes. I tested it in big data. The previous version is using double and it runs slow. This PR I change to float.

And added:
       if (prob_expMembership[i] == 0)
                prob_expMembership[i] = 0.000000001f;
to avoid prob = 0;

jakebiesinger:
> The previous version is using double and it runs slow. 

and changing to float speeds it up?  You sure about that?  Can you show me timings cause flat vs double should have basically 0 overhead difference.  doubles (not floats) are java's native decimal type.

anbangx:
I don't keep the timing. Why do we need double? Float saves space compared
with double.

On Tue, Dec 17, 2013 at 9:09 PM, Jake Biesinger notifications@github.comwrote:

> The previous version is using double and it runs slow.
> 
> and changing to float speeds it up? You sure about that? Can you show me
> timings cause flat vs double should have basically 0 overhead difference.
> doubles (not floats) are java's native decimal type.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/uci-cbcl/genomix/pull/109#issuecomment-30817615
> .

## 

Best Regards,

Anbang Xu

jakebiesinger:
> I don't keep the timing. Why do we need double? Float saves space compared
> with double.

float loses precision compared to double.  In probability, space is not a problem.  precision is.

anbangx:
OK. I will change it to double.

On Tue, Dec 17, 2013 at 9:15 PM, Jake Biesinger notifications@github.comwrote:

> I don't keep the timing. Why do we need double? Float saves space compared
> with double.
> 
> float loses precision compared to double. In probability, space is not a
> problem. precision is.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/uci-cbcl/genomix/pull/109#issuecomment-30817800
> .

## 

Best Regards,

Anbang Xu

anbangx:
But do you think the timing is not different between float and double?

On Tue, Dec 17, 2013 at 9:16 PM, Anbang Xu anbangx@gmail.com wrote:

> OK. I will change it to double.
> 
> On Tue, Dec 17, 2013 at 9:15 PM, Jake Biesinger notifications@github.comwrote:
> 
> > I don't keep the timing. Why do we need double? Float saves space compared
> > with double.
> > 
> > float loses precision compared to double. In probability, space is not a
> > problem. precision is.
> > 
> > —
> > Reply to this email directly or view it on GitHubhttps://github.com/uci-cbcl/genomix/pull/109#issuecomment-30817800
> > .
> 
> ## 
> 
> Best Regards,
> 
> Anbang Xu

## 

Best Regards,

Anbang Xu

jakebiesinger:
> But do you think the timing is not different between float and double?

No, I'm sure it's not.  Especially on a 64-bit host.  You'd have to do some ridiculous operations to see even a tiny difference.

anbangx:
Change float back to double.

jakebiesinger:
LGTM

PR #111, opened 2013-12-18
Implement isStartSeed

anbangx:
@jakebiesinger  Implemented isStartSeed() in RayVertex.

I also suggest to merge Ray branch with main branch as soon as possible.

### DIFF
--- a/genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/config/GenomixJobConf.java
+++ b/genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/config/GenomixJobConf.java
@@ -39,7 +39,7 @@ public class GenomixJobConf extends JobConf {
 
     public static boolean debug;
     public static ArrayList<VKmer> debugKmers;
-    
+
     public static ArrayList<Integer> readLengths;
     public static HashMap<Byte, Integer> outerDistanceMeans;
     public static HashMap<Byte, Integer> outerDistanceStdDevs;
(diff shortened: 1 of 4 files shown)


jakebiesinger:
I was thinking last night... I think it would be cool to have several criteria here instead of length.  Tips almost always have quite long merged nodes but make terrible seeds.  I think it would be slick to do something like:

``` python
minCoverage, maxCoverage = fittedNormalMean - fittedNormalStd, fittedNormalMean + fittedNormalStd
candidates = [v for v in vertices if minCoverage < v.coverage < maxCoverage]
for c in candidates:
    c.seedScore = v.kmerLength * v.numReads
candidates.sort(key=seedScore)
top100 = candidates[-100:]
```

jakebiesinger:
LGTM for now. the idea above can wait.

anbangx:
It's also easy to do. How about add an command line option for user to choose?

jakebiesinger:
> How about add an command line option for user to choose?

I guess so.  More like "a command line option so we can try it out".  I hope the user never, ever has to see anything like this. Maybe submit a separate PR and go ahead and merge this one in?

anbangx:
1. Change namespaces for minLength in conf
2. Add normalMean and normalStd to conf

@jakebiesinger You can merge with this.

jakebiesinger:
The static value usage is pretty awkward.  What if I want to do the fitting again on some subset of the data?

PR #114, opened 2013-12-18
test readHeadoffset after randomly merging

Nan-Zhang:
@jakebiesinger @JavierJia @anbangx @Elmira88 please have a look. complete a pipeline test to test whether readHeadOffset still keep correct offset after our P4 random merging. 
And I already add expected results automatic creating function to compare our actual results, now the test is passed.


jakebiesinger:
whoa which branch is this based on, @Nan-Zhang?  It's pulled in tons of commits here and I can't tell which is which.  At the top of this page, you should specify which branch you based on... or rebase only the relevant changes on top of the main branch.

PR #115, opened 2013-12-19
test corrrectness of readoffset using pipeline tests

Nan-Zhang:
@jakebiesinger I used your instruction git cherry-pick to pick only my changes on top of our main branch :). 
@jakebiesinger @JavierJia @anbangx @Elmira88 please have a look. complete a pipeline test to test whether readHeadOffset still keep correct offset after our P4 random merging. 
And I already add expected results automatic creating function to compare our actual results, now the test is passed.


jakebiesinger:
Hi Nan, can you describe what you're actually testing here?  I worked a fair bit today on the Ray code and it still seems to me that there's some error in offsets.

JavierJia:
merged by #121

PR #116, opened 2013-12-20
Fix some bugs in order to run in the big data

anbangx:
@JavierJia @jakebiesinger @Elmira88 @Nan-Zhang 

When we try to run in the big data, there are some bugs. This PR fixed them. Have a look!

### DIFF
--- a/genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ExternalableTreeSet.java
+++ b/genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ExternalableTreeSet.java
@@ -29,7 +29,8 @@ public abstract class ExternalableTreeSet<T extends WritableComparable<T> & Seri
      */
     private static final long serialVersionUID = 1L;
     static FileManager manager;
-    static private int countLimit = Integer.MAX_VALUE;
+//    static private int countLimit = Integer.MAX_VALUE;
+    static private int countLimit = 1000;
 
     public static synchronized void setupManager(Configuration conf, Path workPath) throws IOException {
         if (manager == null) {
@@ -253,6 +254,7 @@ public abstract class ExternalableTreeSet<T extends WritableComparable<T> & Seri
             } else if (isChanged) {
                 save(path, inMemorySet);
             }
+            System.err.println("Write to HDFS:" + path);
             out.writeUTF(path.toString());
         }
         isChanged = false;
--- a/genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
+++ b/genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
@@ -198,8 +198,8 @@ public class GenomixDriver {
             case MERGE:
             case MERGE_P4:
                 pregelixJobs.add(P4ForPathMergeVertex.getConfiguredJob(conf, P4ForPathMergeVertex.class));
-                flushPendingJobs(conf);
-                setMinScaffoldingSeedLength(conf);
+               // flushPendingJobs(conf);
+               // setMinScaffoldingSeedLength(conf);
                 break;
             case UNROLL_TANDEM:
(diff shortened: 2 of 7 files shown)


jakebiesinger:
It seems the only differences are to set the max memory for the stats job
like I was saying and to set the read id set to actually dump to disk (I
think this patch is already in one of the branches?)

Or was the problem in generating the histogram for the sorted counter min
length?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ExternalableTreeSet.java
```diff
@@ -29,7 +29,8 @@
      */
     private static final long serialVersionUID = 1L;
     static FileManager manager;
-    static private int countLimit = Integer.MAX_VALUE;
+//    static private int countLimit = Integer.MAX_VALUE;
+    static private int countLimit = 1000;
```
JavierJia, line 33: we should remove this debug line. and recover the original commented out line.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ExternalableTreeSet.java
```diff
@@ -253,6 +254,7 @@ public void write(DataOutput out) throws IOException {
             } else if (isChanged) {
                 save(path, inMemorySet);
             }
+            System.err.println("Write to HDFS:" + path);
```
JavierJia, line 257: remove this debug message.

genomix/genomix-driver/src/main/java/edu/uci/ics/genomix/driver/GenomixDriver.java
```diff
@@ -198,8 +198,8 @@ private void addStep(GenomixJobConf conf, Patterns step) throws Exception {
             case MERGE:
             case MERGE_P4:
                 pregelixJobs.add(P4ForPathMergeVertex.getConfiguredJob(conf, P4ForPathMergeVertex.class));
-                flushPendingJobs(conf);
-                setMinScaffoldingSeedLength(conf);
+               // flushPendingJobs(conf);
+               // setMinScaffoldingSeedLength(conf);
```
JavierJia, line 202: Recover these two lines.

genomix/genomix-hadoop/src/main/java/edu/uci/ics/genomix/hadoop/utils/GraphStatistics.java
```diff
@@ -73,6 +80,7 @@ public void map(VKmer key, Node value, OutputCollector<Text, LongWritable> outpu
             throws IOException {
 
         this.reporter = reporter;
+        System.err.println("Inside the Map");
```
JavierJia, line 83: remove this debug line

genomix/genomix-hadoop/src/main/java/edu/uci/ics/genomix/hadoop/utils/GraphStatistics.java
```diff
@@ -130,6 +138,7 @@ public static Counters run(String inputPath, String outputPath, JobConf baseConf
         JobConf conf = new JobConf(baseConf);
         conf.setJobName("Graph Statistics");
         conf.setMapperClass(GraphStatistics.class);
+        conf.set("mapred.child.java.opts", "-Xmx50g");
```
JavierJia, line 141: remove this debug line.

JavierJia:
@jakebiesinger , the problem was to set the count limit and initialize the manager before each job. 
Otherwise the default countLimit of `ReadHeadSet` is `Interger.Max`. Then every readFields method will load from the frame instead of the hdfs files. 

We call the `setGlobalStaticConstants` when the job initializing.

JavierJia:
Hi, @anbangx , I just saw this branch merged, but those comments didn't cover ?

anbangx:
@JavierJia I changed them based on your comments already. We tried to run as soon as possible, so I didn't send it again. Except this: 
+//    static private int countLimit = Integer.MAX_
-    static private int countLimit = 1000;

PR #118, opened 2013-12-26
Wbiesing/readheadset uses static configuration

jakebiesinger:
@anbangx @JavierJia 

We can't rely on any transient member variables staying intact since all classes are created via reflection and `newInstance`.  As a result, the `forceReadEntireBody` and `write*` changes were being lost.

This PR changes those values in two ways: 
- puts the `readEntireBody` flag into the `readFields` stream rather than some configuration.  Otherwise, it's impossible to know how the _previous_ `write` stored the ReadHeadSet.
- makes the `writeEntireBody` a static configuration parameter which is now set inside of the BUILD_HYRACKS code and in the pregelix vertex input and output format adapters.

### DIFF
--- a/genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/config/GenomixJobConf.java
+++ b/genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/config/GenomixJobConf.java
@@ -622,7 +622,7 @@ public class GenomixJobConf extends JobConf {
     public static void setGlobalStaticConstants(Configuration conf) throws IOException {
         Kmer.setGlobalKmerLength(Integer.parseInt(conf.get(KMER_LENGTH)));
         //        ExternalableTreeSet.setupManager(conf, new Path(conf.get("hadoop.tmp.dir", "/tmp")));
-        ExternalableTreeSet.setupManager(conf, new Path("/user/tmp"));
+        ExternalableTreeSet.setupManager(conf, new Path("tmp"));
         ExternalableTreeSet.setCountLimit(1000);
 
         if (conf.get(READ_LENGTHS) != null) {
--- a/genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ExternalableTreeSet.java
+++ b/genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ExternalableTreeSet.java
@@ -55,8 +55,7 @@ public abstract class ExternalableTreeSet<T extends WritableComparable<T> & Seri
     protected boolean isLoaded;
     protected boolean writeToLocal;
     protected boolean readFromLocal;
-    private boolean writeEntireBody;
-    private boolean readEntireBody;
+    private static boolean writeEntireBody;
 
     public ExternalableTreeSet() {
         this(false);
@@ -73,7 +72,6 @@ public abstract class ExternalableTreeSet<T extends WritableComparable<T> & Seri
         isLoaded = false;
         this.writeToLocal = writeTolocal;
         this.readFromLocal = this.writeToLocal;
-        this.writeEntireBody = false;
     }
 
     /**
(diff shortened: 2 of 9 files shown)


genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/config/GenomixJobConf.java
```diff
@@ -622,7 +622,7 @@ public static long tock(String counter) {
     public static void setGlobalStaticConstants(Configuration conf) throws IOException {
         Kmer.setGlobalKmerLength(Integer.parseInt(conf.get(KMER_LENGTH)));
         //        ExternalableTreeSet.setupManager(conf, new Path(conf.get("hadoop.tmp.dir", "/tmp")));
-        ExternalableTreeSet.setupManager(conf, new Path("/user/tmp"));
+        ExternalableTreeSet.setupManager(conf, new Path("tmp"));
```
JavierJia, line 625: Does this work ? I think the hdfs doesn't have the idea of `pwd`, then this "tmp" will not be found. Should it be `/tmp`?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ExternalableTreeSet.java
```diff
@@ -247,14 +246,18 @@ public void readFields(DataInput in) throws IOException {
 
     @Override
     public void write(DataOutput out) throws IOException {
-        out.writeInt(inMemorySet.size());
-        if (!writeEntireBody && !isLoaded && path != null && readFromLocal == writeToLocal) {
+        boolean pathsUnmodified = !writeEntireBody && !isLoaded && path != null && readFromLocal == writeToLocal;
```
JavierJia, line 249: why here is `!isLoad` ? Should it be `!isChanged` ?

genomix/genomix-data/src/test/java/edu/uci/ics/genomix/data/types/ExternalableTreeSetTest.java
```diff
@@ -244,6 +245,7 @@ public void TestWriteEntireBody() {
         ExternalableTreeSet.setCountLimit(limit);
 
         try {
+            ExternalableTreeSet.manager = null;
```
JavierJia, line 248: Why we need to set the `manager =null` ?

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/config/GenomixJobConf.java
jakebiesinger, line 625: in hdfs, the "working dir" is the user's home directory.  I'm not sure what the right place is for everyone but `/user/tmp` gives me permission problems when using `runLocal`

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ExternalableTreeSet.java
jakebiesinger, line 249: I just copied the logic... you mean you think this was a bug from your code?

genomix/genomix-data/src/test/java/edu/uci/ics/genomix/data/types/ExternalableTreeSetTest.java
jakebiesinger, line 248: the manager is static and the constructor for ExternalableTreeSet only created a manager (and changed its path) if one hadn't been created. so this nulling out just pretends the manager class hasn't been used before.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ExternalableTreeSet.java
JavierJia, line 249: I think it should be my bug. `!isLoaded` is more strict than `!isChanged`.

PR #119, opened 2013-12-27
change fs.create everytime to avoid the fs.close exception

JavierJia:
@jakebiesinger @anbangx 
This PR is to creat a DataOutputStream each time, instead of using the old one. This might be the problems of the `Stream Closed` exceptions. And this PR is based on the `wbiesing/readheadset-uses-static-configuration` branch.


JavierJia:
merged by #121

PR #121, opened 2014-01-04
Wbiesing/3 jan implement ray mutual overlap and long candidates

jakebiesinger:

Hi all,

This branch pulls in #115 #118, #119, and #120  and then adds the mutual overlap and long-candidate search as I mentioned today.  If you want only the diff of the scaffolding changes, have a look at 8b2d00c.

There might be a few small problems with offsets being exactly right and paired-end reads definitely need looking into.  I'm submitting this PR now so we can all look at the code and I'd be happy to see this branch merged into the main branch-- it supersedes all my other branches and since it includes multiple PR's in it, it probably supersedes some of your own :)

Please make a close examination and let's make sure I didn't make dumb mistakes :stuck_out_tongue: 

Thanks, and Happy New Year!


jakebiesinger:
I just added from 70ae44f onward which fixes two bugs I found testing last night.  I think that FORWARD, single-end reads with a single seed should work perfectly now :+1:.  We'll need to think about how to handle multiple seeds a bit more carefully-- they are likely to intersect with each other much more frequently now that we do search in the graph.

For the synthetic data with genome length 1000 and read length 100, searching in just the `FORWARD` direction led to a final length of 668 (hooray for 2/3 of the genome in one direction!) and 803 iterations(!).  There were 47 "dominant" edges found and 41 path nodes.

I'll do a few runs on the ecoli data today and will report back to Xiaohui.

JavierJia:
:+1:

jakebiesinger:
Just committed (0603149) one hack to alleviate the parallel problem:  just run each seed as a completely separate job, starting at the longest seeds and going until you reach some threshold.  Have a look!

JavierJia:
@jakebiesinger , we plan to merge this branch first , and then work on the test cases to get more information from the code and also the optimizations.

PR #129, opened 2014-01-21
Jianfeng/shift remove lowcoverage

JavierJia:
@Elmira88 @anbangx 
@jakebiesinger maybe ? 

I fix the remove lowcoverage issues that remove the head kmer may lose the readSet. I passed the ReadSet to the neighbors to keep the original information. 

@jakebiesinger , in order to use negative offset, I did some changes to the offset bitcode. Specificlly, I use the 24th bit as the sign position, the package and unpackage method need to check the sign bit to set or return the offset field

@Elmira88 , if possible, you can also use this branch to test the new graph to see if the result may change ?


genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ReadHeadInfo.java
```diff
@@ -110,10 +112,16 @@ public static long makeUUID(byte mateId, byte libraryId, long readId, int offset
...
-        if (offset != (offset & ~(-1l << (totalBits - bitsForOffset))))
+        // for the offset, in order to allow the negative offset, we need to do some reformats
+        // the bitsForOffset is split into [sign, rest(23]
+        if (Math.abs(offset) > signMaskForOffset - 1) {
             throw new IllegalArgumentException(
                     "byte specified for offset will lose some of its bits when saved! (was: " + offset
                             + " but only allowed " + bitsForOffset + " bits!");
+        }
+        if (offset < 0) {
+            offset = -offset | signMaskForOffset;
```
jakebiesinger, line 123: So the value is stored as a positive number with an additional sign bit, yes?  (Is that technically 1's complement?)

genomix/genomix-data/src/test/java/edu/uci/ics/genomix/data/types/ReadHeadInfoTest.java
```diff
@@ -33,4 +33,12 @@ public void TestInitial() {
             //Assert.assertEquals(pos1.toString(), pos.toString());
         }
     }
+
+    @Test
+    public void TestNegative() {
```
jakebiesinger, line 38: Can you add failure assertions for items that go out of range in positive and negative directions?

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/removelowcoverage/ShiftLowCoverageReadSetVertex.java
```diff
@@ -0,0 +1,135 @@
...
+    private void clearupMyReadSet() {
+        getVertexValue().getUnflippedReadIds().reset();
+        getVertexValue().getFlippedReadIds().reset();
+    }
+
+    private void shiftReadSetToNeighors() {
+        ReadHeadSet unflippedSet = getVertexValue().getUnflippedReadIds();
+        KmerFactory kmerFactory = new KmerFactory(Kmer.getKmerLength());
+        for (ReadHeadInfo info : unflippedSet) {
+            VKmer destForward = kmerFactory.getSubKmerFromChain(-info.getOffset() + 1, Kmer.getKmerLength(),
```
jakebiesinger, line 67: Which node is the destination here?  I'm not sure what node is pointed to by `-offset + 1`.  In an unmerged graph, all the `ReadHeadInfo.offset`s will be 0 so you're just shifting towards the node's FF or FR neighbor.  If that neighbor is also low coverage, then you'll pass to `-offset + 1` again... wouldn't that be `-1 + 1 = 0` so you're just passing it back to its origin (offset=0)?  Why not just use the node's neighbor list and pass the reads along until you get to a node that has higher coverage?

In a merged graph, the kmers in the read may not have nodes in the graph anymore.  Since we're just doing this step after graphbuild, it should be okay but you might want a disclaimer in the header comments.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/removelowcoverage/ShiftLowCoverageReadSetVertex.java
```diff
@@ -0,0 +1,135 @@
...
+            }
+
+        }
+
+        ReadHeadSet flippedSet = getVertexValue().getFlippedReadIds();
+        for (ReadHeadInfo info : flippedSet) {
+            // <--     
+            //    ----->
+            VKmer destForward = kmerFactory.getSubKmerFromChain(info.getOffset() - (Kmer.getKmerLength() - 1) + 1,
+                    Kmer.getKmerLength(), info.getThisReadSequence());
```
jakebiesinger, line 101: the read that is stored in the ReadHeadInfo object isn't ever flipped.  That is, the sequence is never put in as a reverse-complement.  That means the `flipped` logic is the opposite of what you have in your comments here.  "Flipped" means flipped with respect to the node that is containing the read.

``` java
if (flipped) {
    // <--     2
    //   -->   3
```

jakebiesinger, line 101: Nevermind-- I think you're right.  This is the `flippedReadIds` set which started out `-->` and now the destintation node might have flipped wrt the read.  I think it's okay.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/removelowcoverage/ShiftLowCoverageReadSetVertex.java
JavierJia, line 67: This one should work in the unmerged graph, directly after the graph building. 

My assumption here is that if the offset come from the `unflippedset`, then the `offset` should always `<=0`. Because the direction is the same, and the only shift sequence allowed should be `-1 --> -2 --> -3 --> ...`. And the correspond `flippedset` should always be `>=` `kmerlength-1`. 

I was using the neighborlist before, then I found it really tedious to play with `FF, FR ...`, and also I need to compare the dest with every info inside the ReadHeadSet to get the proper ones. Then I decide to change to directly loop through the `ReadHeadSet` which is more deterministic.

genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ReadHeadInfo.java
JavierJia, line 123: Yes.
:sob: I'm not sure whether it is the 1's complement.

genomix/genomix-data/src/test/java/edu/uci/ics/genomix/data/types/ReadHeadInfoTest.java
JavierJia, line 38: done.

genomix/genomix-pregelix/src/main/java/edu/uci/ics/genomix/pregelix/operator/removelowcoverage/ShiftLowCoverageReadSetVertex.java
jakebiesinger, line 67: Yeah, that's fair enough.  And actually, if there are 2+ neighbors, you
have to make a decision about which to send to, for which you need to look
at the read anyway.

## 

Jake Biesinger
Graduate Student
Xie Lab, UC Irvine

JavierJia, line 67: So we can merge this branch now ?

jakebiesinger, line 67: I haven't thought through what the negative offsets will mean for scoring.
 I recently tried to run the pipeline and found some negative offsets
(which caused errors!).  Actually, there shouldn't be any negative offsets
in the main branch, right?  I'll have to spend a bit more time debugging.

But anyway, yes I think this can merge but maybe make it an option to
include this behavior?

## 

Jake Biesinger
Graduate Student
Xie Lab, UC Irvine

JavierJia, line 67: OK, good suggestion. 
I split the `LOW_COVERAGE` into `SHIFT_LOW_COVERAGE` and `REMOVE_LOW_COVERAAGE`. They are two different jobs. 
The `REMOVE_LOW_COVERAAGE` is the original one, you can run it without shift the previous' readset.
And the `SHIFT_LOW_COVERAGE` is just shifting the ReadSet, without removing the node, so the newer LOW_COVERAGE should be `SHIFT_LOW_COVERAGE,REMOVE_LOW_COVERAGE`.

FYI @Elmira88  also.

PR #143, opened 2014-04-09
Fix validation map been cleaned up issue

JavierJia:
@jakebiesinger I think the problem is fixed now. That test data set run to the end on ipubmed2.


jakebiesinger:
Looks like it fixes the problem. LGTM.

PR #145, opened 2014-04-09
Scaffolding options are now commandline options.

jakebiesinger:
If you want to override the defaults, you have to set the values explicitly like so:

-scaffold_removeOtherOutgoing true -scaffold_removeOtherIncoming true -scaffold_candidatesScoreWalk true -scaffold_delayPrune true -scaffold_earlyStop false -scaffold_expandCandidateBranchMaxDistance 20


Elmira88:
LGTM! :)

PR #149, opened 2014-04-22
Wbiesing/bubble search removes nodes in path

jakebiesinger:
@Elmira88


jakebiesinger:
Sorry this was against the wrong base branch.  See #150

PR #151, opened 2014-04-23
remove autodelete in ExternalTreeSet

JavierJia:
@jakebiesinger Here is the simple change. It should solve the problem temporary.

### DIFF
--- a/genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ExternalableTreeSet.java
+++ b/genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ExternalableTreeSet.java
@@ -249,10 +249,13 @@ public abstract class ExternalableTreeSet<T extends WritableComparable<T> & Seri
             for (T t : inMemorySet) {
                 writeNonGenericElement(out, t);
             }
-            if (path != null) {
-                FileManager.getManager().deleteFile(path, writeToLocal);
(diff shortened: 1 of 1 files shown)


genomix/genomix-data/src/main/java/edu/uci/ics/genomix/data/types/ExternalableTreeSet.java
```diff
@@ -249,10 +249,13 @@ public void write(DataOutput out) throws IOException {
...
-            if (path != null) {
-                FileManager.getManager().deleteFile(path, writeToLocal);
-                path = null;
-            }
+            /**
+             * Temporary solution to remove this delete,
+             */
+            //            if (path != null) {
+            //                FileManager.getManager().deleteFile(path, writeToLocal);
+            //                path = null;
```
jakebiesinger, line 257: You should still call `path = null;`... just don't delete the file, yes?

JavierJia, line 257: I was thinking that before. Because the file is still there, thus the path is still valid. So after `write` if someone still need this path, it can still obtain (or duplicate) from this path ? 
And also because the auto `delete` is removed, then it should be self maintained by the `ExternalTreeSet`. Once he want to delete the path, he can use this variable to explicit delete it.

